PR #34, opened 2020-02-02
Added `environment` to CreateWorkflow request

Vafilor:


api/workflow.proto
```diff
@@ -123,10 +123,11 @@ message Workflow {
     string uid = 2;
     string name = 3;
     string status = 4;
+    bool environment = 5;
```
rushtehrani: @Vafilor can you elaborate on the reason for this field?

rushtehrani: just realized this is for environment variables. I’m thinking about automatically associating environment variables to every container node, and removing this from both UI and here, thoughts?

rushtehrani:
Closing this based on our discussion.

PR #35, opened 2020-02-03
Feature/GitHub.core.24 secret.management.mounting

aleksandrmelnikov:


rushtehrani:
@aleksandrmelnikov see the conflicts, it has to do with the changes I mentioned yesterday.

rushtehrani:
test

kube/workflow.go
```diff
@@ -101,6 +103,50 @@ func (c *Client) create(namespace string, wf *Workflow, opts *WorkflowOptions) (
 		}
 	}
 
+	secretName := "onepanel-default-env"
```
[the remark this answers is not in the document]
aleksandrmelnikov: Hmm, how would I get the secret name from inside Core?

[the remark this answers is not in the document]
rushtehrani: @aleksandrmelnikov basically the client would send `onepanel-default-env` as the secret name instead of having it be hardcoded in the API.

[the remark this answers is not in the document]
rushtehrani: @aleksandrmelnikov my bad, I just realized this is in the workflow creation and not the secret API.

kube/workflow.go
```diff
@@ -101,6 +103,50 @@ func (c *Client) create(namespace string, wf *Workflow, opts *WorkflowOptions) (
...
+	addSecretValsToTemplate := true
+	secret, secretGetErr := c.Clientset.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
```
rushtehrani: I think it might be cleaner to use `err` here so it's not redefined again and then below it instead of:
```
if err, ok := secretGetErr.(*errors.StatusError); ok
```
Use the new `As` method:
```
var statusError *errors.StatusError
if errors.As(err, &statusError) {
```

kube/workflow.go
```diff
@@ -6,6 +6,8 @@ import (
 	"github.com/argoproj/argo/workflow/common"
 	"github.com/argoproj/pkg/json"
 	"github.com/onepanelio/core/util/env"
+	v12 "k8s.io/api/core/v1"
```
rushtehrani: Let's keep calling these `corev1` for consistency.

kube/workflow.go
```diff
@@ -6,6 +6,8 @@ import (
...
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
rushtehrani: Just realized `"k8s.io/apimachinery/pkg/apis/meta/v1"` was accidentally redefined twice.

kube/workflow.go
aleksandrmelnikov: Neat to know about `As`!

manager/secret_manager.go
```diff
@@ -8,6 +8,30 @@ func (r *ResourceManager) CreateSecret(namespace string, secret *model.Secret) (
 	return r.kubeClient.CreateSecret(namespace, secret)
 }
 
-func (r *ResourceManager) GetSecret(namespace, name string) (secret *model.Secret, err error) {
-	return r.kubeClient.GetSecret(namespace, name)
+func (r *ResourceManager) SecretExists(namespace string, secretName string) (exists bool, err error) {
+	return r.kubeClient.SecretExists(namespace, secretName)
+}
+
+func (r *ResourceManager) GetSecret(namespace string, secret *model.Secret) (secretRes *model.Secret, err error) {
```
rushtehrani: Shouldn't this just take name? 
```
func (r *ResourceManager) GetSecret(namespace, name string) (secret *model.Secret, err error) {
```

manager/secret_manager.go
```diff
@@ -8,6 +8,30 @@ func (r *ResourceManager) CreateSecret(namespace string, secret *model.Secret) (
...
+	return r.kubeClient.GetSecret(namespace, secret.Name)
+}
+
+func (r *ResourceManager) ListSecrets(namespace string) (secrets []model.Secret, err error) {
```
rushtehrani: The return var should be `secrets []*model.Secret`

rushtehrani:
@aleksandrmelnikov did you rebuild the proto files after rebase?

kube/secret.go
```diff
@@ -17,20 +22,210 @@ func (c *Client) CreateSecret(namespace string, secret *model.Secret) (err error
 	return
 }
 
-func (c *Client) GetSecret(namespace, name string) (secret *model.Secret, err error) {
-	s, err := c.CoreV1().Secrets(namespace).Get(name, metav1.GetOptions{})
+func (c *Client) SecretExists(namespace string, secretName string) (exists bool, err error) {
```
rushtehrani: `secretName` => `name` for consistency

manager/secret_manager.go
```diff
@@ -8,6 +8,30 @@ func (r *ResourceManager) CreateSecret(namespace string, secret *model.Secret) (
 	return r.kubeClient.CreateSecret(namespace, secret)
 }
 
-func (r *ResourceManager) GetSecret(namespace, name string) (secret *model.Secret, err error) {
-	return r.kubeClient.GetSecret(namespace, name)
+func (r *ResourceManager) SecretExists(namespace string, secretName string) (exists bool, err error) {
```
rushtehrani: `secretName` => `name` for consistency

aleksandrmelnikov:
> @aleksandrmelnikov did you rebuild the proto files after rebase?

Yes

kube/secret.go
aleksandrmelnikov: Refactored.

manager/secret_manager.go
aleksandrmelnikov: I'll do this change, but why use []*?

manager/secret_manager.go
aleksandrmelnikov: Refactored.

manager/secret_manager.go
aleksandrmelnikov: How does the API know what namespace to use?

manager/secret_manager.go
rushtehrani: Good question, it's possible these secrets can go up to 1MB (let's say if there is a cert or multiple certs in them). In this case you're not copying each struct if you iterate through the slice or pass the slice from method to another.

manager/secret_manager.go
rushtehrani: You'd still pass in the `namespace`,  meant shouldn't it take `name` instead of `*model.Secret`? Like this:
```
func (r *ResourceManager) GetSecret(namespace, name string) (secret *model.Secret, err error)
```

manager/secret_manager.go
aleksandrmelnikov: Interesting, good to know.

manager/secret_manager.go
aleksandrmelnikov: I think I was following the CreateSecret for refactoring.. It seems like CreateSecret converts to a model inside the `secret_server.go`, instead of the `secret_manager.go`.
To verify, I should be creating the model inside the `secret_manager.go`?
And refactor CreateSecret inside of secret_server.go?

manager/secret_manager.go
```diff
@@ -41,7 +41,19 @@ func (r *ResourceManager) AddSecretKeyValue(namespace string, name string, key s
 	return r.kubeClient.AddSecretKeyValue(namespace, name, key, value)
 }
 
-func (r *ResourceManager) UpdateSecretKeyValue(namespace string, name string, key string, value string) (updated bool, err error) {
+func (r *ResourceManager) UpdateSecretKeyValue(namespace string, name string, data map[string]string) (updated bool, err error) {
```
rushtehrani: I am thinking this should just take `*model.Secret` as a param instead of separate arguments `name string, data map[string]string`

aleksandrmelnikov: Good point, refactoring.

manager/secret_manager.go
```diff
@@ -41,21 +41,21 @@ func (r *ResourceManager) AddSecretKeyValue(namespace string, name string, key s
 	return r.kubeClient.AddSecretKeyValue(namespace, name, key, value)
 }
 
-func (r *ResourceManager) UpdateSecretKeyValue(namespace string, name string, data map[string]string) (updated bool, err error) {
-	if len(data) == 0 {
+func (r *ResourceManager) UpdateSecretKeyValue(namespace string, secret model.Secret) (updated bool, err error) {
```
rushtehrani: this should be `*model.Secret` so we pass by ref.

manager/secret_manager.go
```diff
@@ -1,13 +1,114 @@
...
+
+func (r *ResourceManager) DeleteSecretKey(namespace string, name string, key string) (deleted bool, err error) {
+	return r.kubeClient.DeleteSecretKey(namespace, name, key)
+}
+
+func (r *ResourceManager) AddSecretKeyValue(namespace string, name string, key string, value string) (inserted bool, err error) {
+	return r.kubeClient.AddSecretKeyValue(namespace, name, key, value)
+}
+
+func (r *ResourceManager) UpdateSecretKeyValue(namespace string, secret *model.Secret) (updated bool, err error) {
```
aleksandrmelnikov: @rushtehrani Should the function param here be a model.Secret? Doesn't this mean that the Server is creating the model, instead of passing the data along?
Shouldn't the manager be creating the model?

rushtehrani: @aleksandrmelnikov no, this is correct, the Server is responsible for converting the api request message to a model and pass into manager.

aleksandrmelnikov: Hmm, okay. So it's okay if the manager takes a model, passes a model, returns a model?

rushtehrani: Yes

PR #38, opened 2020-02-04
Support archiving workflow templates

rushtehrani:
Request syntax:
```
 DELETE http://<host>/apis/v1beta1/<namespace>/workflow_templates/<uid>
```


manager/workflow_manager.go
```diff
@@ -346,3 +346,28 @@ func (r *ResourceManager) ListWorkflowTemplates(namespace string) (workflowTempl
...
+	}
+	if !allowed {
+		return false, util.NewUserError(codes.PermissionDenied, "Permission denied.")
+	}
+
+	workflowTemplate, err := r.workflowRepository.GetWorkflowTemplate(namespace, uid, 0)
+	if err != nil {
+		return false, util.NewUserError(codes.Unknown, "Unknown error.")
+	}
+	if err == nil && workflowTemplate == nil {
```
Vafilor: No need to check if err == nil - it was always be nil here because of the above check.

rushtehrani: Updated this and other instances of this check.

PR #53, opened 2020-02-10
Feature/add.logging.with.logrus

aleksandrmelnikov:


manager/secret_manager.go
```diff
@@ -21,14 +27,30 @@ func (r *ResourceManager) CreateSecret(namespace string, secret *model.Secret) (
 func (r *ResourceManager) SecretExists(namespace string, name string) (exists bool, err error) {
 	var foundSecret *model.Secret
 	var statusError *errors.StatusError
+
 	foundSecret, err = r.kubeClient.SecretExists(namespace, name)
 	if err != nil {
 		if goerrors.As(err, &statusError) {
 			if statusError.ErrStatus.Reason == "NotFound" {
+				logging.Logger.Log.WithFields(log.Fields{
```
rushtehrani: @aleksandrmelnikov I don't think we need to log status error like separately, I think logging just the top most error would be sufficient.

manager/secret_manager.go
```diff
@@ -21,14 +27,30 @@ func (r *ResourceManager) CreateSecret(namespace string, secret *model.Secret) (
...
 				return false, util.NewUserError(codes.NotFound, "Secret Not Found.")
 			}
+			logging.Logger.Log.WithFields(log.Fields{
+				"Namespace": namespace,
+				"Name":      name,
+				"Error":     err.Error(),
+			}).Error("Checking existence of secret.")
 			return false, util.NewUserError(codes.Unknown, "Error when checking existence of secret.")
 		}
+		logging.Logger.Log.WithFields(log.Fields{
```
rushtehrani: So basically move this up right under `if err !== nil` and log `err` once. Thoughts?

aleksandrmelnikov: Hmm, I think I see your point. Moving it up right under err !== nil, would capture the error.
The rest if-else checks are for user reporting.

Should this adjustment be applied elsewhere too?

rushtehrani: That's correct. Yes, we should adjust this everywhere else.

rushtehrani:
@aleksandrmelnikov everything else LGTM, I'll merge once you make the adjustments we discussed.

manager/secret_manager.go
aleksandrmelnikov: Right on, will do

rushtehrani:
@aleksandrmelnikov let's get this refactored so we can merge it before we get more conflicts.

aleksandrmelnikov:
@rushtehrani Refactored per feedback.

PR #57, opened 2020-02-13
Feature/workflow.pagination

Vafilor:


server/workflow_server.go
```diff
@@ -151,19 +155,41 @@ func (s *WorkflowServer) GetWorkflowMetrics(ctx context.Context, req *api.GetWor
...
+		req.Page = 1
+	}
+
 	workflows, err := s.resourceManager.ListWorkflows(req.Namespace, req.WorkflowTemplateUid, req.WorkflowTemplateVersion)
 	if errors.As(err, &userError) {
 		return nil, userError.GRPCError()
 	}
 
-	apiWorkflows := []*api.Workflow{}
+	apiWorkflows := make([]*api.Workflow, 0)
```
rushtehrani: is this more efficient or just preference?

Vafilor: My IDE complains at me when it sees it.

It has:

Inspection info: Reports slice declaration with empty literal initializer instead of nil.
An empty slice can be represented by nil or an empty slice literal. The first approach is preferred as it does not lead to memory allocation.


So it actually would prefer 
```var apiWorkflows []*api.Workflow```

Which does work.

rushtehrani: That is actually much better, we should use `var apiWorkflows []*api.Workflow` so it doesn't allocate any memory.

PR #68, opened 2020-02-21
Rename Workflow to WorkflowExecution

rushtehrani:


api/workflow.proto
```diff
@@ -9,52 +9,52 @@ import "metric.proto";
 
 service WorkflowService {
     // Creates a Workflow
-    rpc CreateWorkflow (CreateWorkflowRequest) returns (Workflow) {
+    rpc CreateWorkflowExecution (CreateWorkflowExecutionRequest) returns (WorkflowExecution) {
         option (google.api.http) = {
-            post: "/apis/v1beta1/{namespace}/workflows"
-            body: "workflow"
+            post: "/apis/v1beta1/{namespace}/workflow_exeuctions"
```
aleksandrmelnikov: @rushtehrani Typo?

rushtehrani: Good catch! there may be more of these.

aleksandrmelnikov: That's the only one I've caught so far, maybe Andrey will see one more :)

aleksandrmelnikov:
Approval overall, just that typo.

api/workflow.proto
rushtehrani: Fixed

PR #125, opened 2020-03-24
fix: Updated authentication check endpoint to take the auth token as …

Vafilor:
…POST request data


rushtehrani:
@Vafilor getting these errors. Also is there a UI update?
```bash
$ go run main.go
# github.com/onepanelio/core/server/auth
server/auth/auth.go:87:18: undefined: errors
server/auth/auth.go:90:30: undefined: api
server/auth/auth.go:92:18: undefined: errors
```

Issue #127, opened 2020-03-24
Clone WorkflowTemplate


rushtehrani:
Need to be able to clone a WorkflowTemplate from UI and API.

aleksandrmelnikov:
Clone details:
- They have to pick a unique new name for the WorkflowTemplate
- Do we copy over the history of versions?
- What about the related WorkflowExecutions? If a WorkflowTemplate had a bunch of Workflows executed, do we want to bring those over as well?
- Other details?

rushtehrani:
> * They have to pick a unique new name for the WorkflowTemplate
Correct

> * Do we copy over the history of versions?
No, only latest template cloned in this phase.

> * What about the related WorkflowExecutions? If a WorkflowTemplate had a bunch of Workflows executed, do we want to bring those over as well?
No, this is a pure template clone in this phase.

PR #131, opened 2020-03-30
Added endpoint to clone workflow templates.

aleksandrmelnikov:
- Added endpoint to proto
- Added supporting functions and error checking

closes https://github.com/onepanelio/core/issues/127


pkg/workflow_template.go
```diff
@@ -178,6 +178,29 @@ func (c *Client) getWorkflowTemplate(namespace, uid string, version int32) (work
 	return
 }
 
+func (c *Client) getWorkflowTemplateByName(namespace, name string, version int32) (workflowTemplate *WorkflowTemplate, err error) {
```
Vafilor, line 181: Why do you need a version if you're getting a template by name?

Vafilor: @aleksandrmelnikov

pkg/workflow_template.go
```diff
@@ -346,6 +369,28 @@ func (c *Client) GetWorkflowTemplate(namespace, uid string, version int32) (work
 	return
 }
 
+func (c *Client) GetWorkflowTemplateByName(namespace, name string, version int32) (workflowTemplate *WorkflowTemplate, err error) {
```
Vafilor, line 372: Why do you need a version if you're getting a template by name @aleksandrmelnikov ?

pkg/workflow_template.go
```diff
@@ -346,6 +369,28 @@ func (c *Client) GetWorkflowTemplate(namespace, uid string, version int32) (work
...
+		return nil, util.NewUserError(codes.Unknown, "Unknown error.")
+	}
+	if workflowTemplate == nil {
+		return nil, util.NewUserError(codes.NotFound, "Workflow template not found.")
+	}
+
+	return
+}
+
+func (c *Client) CloneWorkflowTemplate() (workflowTemplate *WorkflowTemplate, err error) {
```
Vafilor, line 389: Is this supposed to do anything?

pkg/workflow_template.go
aleksandrmelnikov: @Vafilor This chunk of code:
```
	sb := c.workflowTemplatesSelectBuilder(namespace).Where(sq.Eq{"wt.name": name}).
		Columns("wtv.manifest").
		OrderBy("wtv.version desc").
		Limit(1)
	if version != 0 {
		sb = sb.Where(sq.Eq{"wtv.version": version})
	}
```
The SelectBuilder does a join with the workflow_template_versions table.
So there are multiple rows available for a given workflow_template by name, right?
So you can be more explicit with by using version.

Is this line of reasoning correct?

pkg/workflow_template.go
aleksandrmelnikov: Good catch. I ended up implementing the functionality in the Server piece instead of the Client.

pkg/workflow_template.go
Vafilor: Ah, yeah, that makes sense.

Maybe we should add a method later that doesn't take a version or something, but all good.

PR #151, opened 2020-04-13
Feature/core.update.template

rushtehrani:


rushtehrani:
Noticed the naming convention for migrations is changed, best to keep those consistent and match SQL naming conventions.

Vafilor:
> Noticed the naming convention for migrations is changed, best to keep those consistent and match SQL naming conventions.

@rushtehrani Oh interesting. Snake case then?

rushtehrani:
> @rushtehrani Oh interesting. Snake case then?

Yes, noticed we missed a couple from before too.

Vafilor:
> > @rushtehrani Oh interesting. Snake case then?
> 
> Yes, noticed we missed a couple from before too.

@rushtehrani Pleas check the updated file names.

rushtehrani:
> Please check the updated file names.

Great, they're good to go.

rushtehrani:
Looking at WorkflowTemplate names in K8s, possible to use `<workflow-template-name>-<version>` instead of the UUID? Haven't dug in, but how're versions handled currently?

Vafilor:
> Looking at WorkflowTemplate names in K8s, possible to use `<workflow-template-name>-<version>` instead of the UUID? Haven't dug in, but how're versions handled currently?

I think we can do that. Right now versions are a label on each workflow template, along with an additional label indicating if it is the latest version.

The names are just random uuids. 

If we switch to name-version, we should be fine as the workflow template names must be unique across namespaces. 

However, you will no longer be able to create something like this.

Workflow template name: zebra

And then create

zebra-1 or zebra-2, because these names will be taken in k8s.

pkg/util/label/label.go
```diff
@@ -4,6 +4,15 @@ import (
 	"strings"
 )
 
+const (
+	OnepanelPrefix      = "onepanel.io/"
+	TagPrefix           = "tags.onepanel.io/"
+	WorkflowTemplate    = OnepanelPrefix + "workflow_template"
```
rushtehrani, line 10: For consistency, we should always use `-` instead of `_` for label keys and values.

server/workflow_template_server.go
```diff
@@ -0,0 +1,334 @@
...
+
+		result = append(result, keyValue)
+	}
+
+	return result
+}
+
+func (s *WorkflowTemplateServer) CreateWorkflowTemplate(ctx context.Context, req *api.CreateWorkflowTemplateRequest) (*api.WorkflowTemplate, error) {
+	client := ctx.Value("kubeClient").(*v1.Client)
+	allowed, err := auth.IsAuthorized(client, req.Namespace, "create", "argoproj.io", "workflows", "")
```
rushtehrani, line 59: Now that we are using `WorkflowTemplates`, this should check for `workflowtemplates` instead of `workflows`:
```
auth.IsAuthorized(client, req.Namespace, "create", "argoproj.io", "workflowtemplates", "")
```

rushtehrani: This applies to all the other `auth.IsAuthorized` calls in this file.

server/workflow_template_server.go
```diff
@@ -0,0 +1,334 @@
...
+	return &api.ArchiveWorkflowTemplateResponse{
+		WorkflowTemplate: &api.WorkflowTemplate{
+			IsArchived: archived,
+		},
+	}, nil
+}
+
+func (s *WorkflowTemplateServer) GetWorkflowTemplateLabels(ctx context.Context, req *api.GetWorkflowTemplateLabelsRequest) (*api.GetLabelsResponse, error) {
+	client := ctx.Value("kubeClient").(*v1.Client)
+	allowed, err := auth.IsAuthorized(client, req.Namespace, "create", "argoproj.io", "workflows", "")
```
rushtehrani, line 248: This should be `"get"` not `"create"`

server/workflow_template_server.go
```diff
@@ -0,0 +1,334 @@
...
+		Labels: mapToKeyValue(labels),
+	}
+
+	return resp, nil
+}
+
+// Deletes all of the old labels and adds the new ones.
+func (s *WorkflowTemplateServer) ReplaceWorkflowTemplateLabels(ctx context.Context, req *api.ReplaceLabelsRequest) (*api.GetLabelsResponse, error) {
+	client := ctx.Value("kubeClient").(*v1.Client)
+	allowed, err := auth.IsAuthorized(client, req.Namespace, "create", "argoproj.io", "workflows", "")
```
rushtehrani, line 294: I believe the verb here should be `"update"`

server/workflow_template_server.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	return resp, nil
+}
+
+// Adds any labels that are not yet associated to the workflow execution.
+// If the label already exists, overwrites it.
+func (s *WorkflowTemplateServer) AddWorkflowTemplateLabels(ctx context.Context, req *api.AddLabelsRequest) (*api.GetLabelsResponse, error) {
+	client := ctx.Value("kubeClient").(*v1.Client)
+	allowed, err := auth.IsAuthorized(client, req.Namespace, "create", "argoproj.io", "workflows", "")
```
rushtehrani, line 269: I believe the verb here should be `"update"`

server/workflow_template_server.go
```diff
@@ -0,0 +1,334 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return apiWorkflowTemplate(workflowTemplate), nil
+}
+
+func (s *WorkflowTemplateServer) CloneWorkflowTemplate(ctx context.Context, req *api.CloneWorkflowTemplateRequest) (*api.WorkflowTemplate, error) {
+	client := ctx.Value("kubeClient").(*v1.Client)
+	allowed, err := auth.IsAuthorized(client, req.Namespace, "get", "argoproj.io", "workflows", "")
```
rushtehrani, line 141: I think only the `"create"` check is sufficient here since that's the highest permission.

rushtehrani:
Getting this error when using `parameters`:
image

rushtehrani:
Parameters are not rendering in the execution form:
image

rushtehrani:
Looks like empty fields show up after a template is saved. These were never added in the original template:
image

pkg/types.go
```diff
@@ -58,8 +60,10 @@ type WorkflowTemplate struct {
 	Name       string
 	Manifest   string
 	Version    int32
-	IsLatest   bool `db:"is_latest"`
+	IsLatest   bool
 	IsArchived bool `db:"is_archived"`
+	LatestArgo *wfv1.WorkflowTemplate
```
rushtehrani, line 65: Can we use a better field name? Like `ArgoWorkflowTemplate`.

Vafilor: On it.

server/workflow_template_server.go
Vafilor: Fixed.

server/workflow_template_server.go
Vafilor: Fixed.

server/workflow_template_server.go
Vafilor: Should be fixed.

pkg/util/label/label.go
Vafilor: Done.

Vafilor:
> Parameters are not rendering in the execution form:
> image

Should be fixed.

pkg/types.go
Vafilor: Done.

Vafilor:
> Looks like empty fields show up after a template is saved. These were never added in the original template:
> image

I added some logic to filter out empty fields. Let me know how it looks for you.

Vafilor:
> Looking at WorkflowTemplate names in K8s, possible to use `<workflow-template-name>-<version>` instead of the UUID? Haven't dug in, but how're versions handled currently?

This is up.

rushtehrani:
parameter types are not being recognized, for example `action` is `input.radio` and `host` is `input.hidden`:
image

rushtehrani:
I think there is a confusion about this button. It is supposed to toggle between graph render and form render.
image

rushtehrani:
Getting this error when executing the workflow mentioned above:
image

pkg/types.go
```diff
@@ -182,3 +299,61 @@ func FilePathToExtension(path string) string {
 
 	return path[dotIndex+1:]
 }
+
+func WrapSpecInK8s(data []byte) ([]byte, error) {
```
rushtehrani, line 303: This function name is a bit confusing.

Vafilor: Yeah, I'm trying to come up with a better name. 

Since we only store the `spec` part of the manifest, this part wraps it so we have 
```{
   manifest: {},
   spec: {...}
}```

instead of just the `spec` contents.

Issue #144, opened 2020-04-09
Display WorkflowExecution stats on Workflow listing page


rushtehrani:
See https://projects.invisionapp.com/d/main#/console/19217882/405979089/preview

Vafilor:
@rushtehrani Are these stats supposed to be live?

Also, are any of these stats going to be stored in database, or do we need to do a query each time?

rushtehrani:
@Vafilor 
> Are these stats supposed to be live?
No, but maybe we can poll for stats every so often.

> Also, are any of these stats going to be stored in database, or do we need to do a query each time?
I think we need to cache/store them but deferring to the implementer to make that decision after research.

Vafilor:
@rushtehrani 

Polling sounds good.

Regarding cache/store, another possibility is to store the data on the workflow template itself as a label.

Might be more efficient to just get it from the DB, but just throwing up some ideas if we want to keep more to k8s.

rushtehrani:
> Regarding cache/store, another possibility is to store the data on the workflow template itself as a label.

I didn't think of that, this might be the best way actually. Wonder if it makes sense to use [Annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) instead if we're not querying objects based on this data?

Vafilor:
I think it's better to assume we will query for it, it provides us more capability.
If we really want, we can add it to both places.

In addition to your example, we might want 
- Sort by least/most failed, # executions, etc.

aleksandrmelnikov:
Overview of decided implementation:

**We are using a database for storing information.**
- Generating statistics is easier than querying kubernetes workflow information
- We can store records of workflows that are gone from k8s
-- Argo garbage collection can delete workflows
-- CronWorkflows have a limit for how many error/successful workflows they keep around
- We don't have to worry about concurrency or race conditions with data updates / inserts / queries.
-- If we were to store the information in k8s via annotation or label, these might be issues.

**We will push information to the database instead of polling.**
- Polling requires running a process every so often. So we'd need to set up a cron
- We'd also be relying on the time-frame being small enough to capture any workflow execution

**We will attempt to use argo "exit handlers" to push information.**
- These are going to run after the workflow completes or errors out
- Adding a side-car container would be too granular, we are focusing on workflow outcomes

**Regarding deletion of workflows, either by garbage collection or cron workflow settings:**
We're going to test out an archive functionality, from argo.
- For now, we may just have an endpoint to report how many are archived.
[...]

PR #153, opened 2020-04-14
Feat/core.144 workflow.stats

aleksandrmelnikov:
closes https://github.com/onepanelio/core/issues/144


pkg/types.go
```diff
@@ -52,14 +52,34 @@ type CronWorkflow struct {
...
+
+type WorkflowExecutionStatistic struct {
+	ID                 uint64
+	WorkflowTemplateId uint64
+	Name               string
+	Namespace          string
+	//Interface to support null values for timestamps, when scanning from db into structs
+	CreatedAt  interface{} `db:"created_at"`
+	FinishedAt interface{} `db:"finished_at"`
+	FailedAt   interface{} `db:"failed_at"`
```
Vafilor, line 82: @aleksandrmelnikov instead of interface{}, would `*time.Time` work?

aleksandrmelnikov: No, because the db value can be null for finished_at or failed_at.
Squirrel throws an error because it can't assign "nil" to "time.Time"

pkg/workflow_template.go
```diff
@@ -410,6 +410,10 @@ func (c *Client) ListWorkflowTemplates(namespace string) (workflowTemplateVersio
 		return nil, util.NewUserError(codes.NotFound, "Workflow templates not found.")
 	}
 
+	for _, workflowTemplate := range workflowTemplateVersions {
+		err = c.GetWorkflowExecutionStatisticsForTemplate(workflowTemplate)
```
Vafilor, line 414: If there is an error, let's log it.

pkg/types.go
aleksandrmelnikov: Actually, make sense. Pushed up change.

pkg/workflow_template.go
aleksandrmelnikov: Added logging.

pkg/workflow_execution.go
```diff
@@ -327,6 +348,49 @@ func (c *Client) CreateWorkflowExecution(namespace string, workflow *WorkflowExe
...
 
+func (c *Client) AddWorkflowExecutionStatistic(namespace, name string, workflowTemplateID int64, createdAt time.Time, workflowOutcomeIsSuccess bool) (err error) {
+	tx, err := c.DB.Begin()
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
+	if workflowOutcomeIsSuccess {
+		_, err := sb.Insert("workflow_executions").
+			SetMap(sq.Eq{
```
rushtehrani, line 359: Wonder if this can be done a bit more DRY with `sb`. Can't you just change only one key in the map?

pkg/workflow_execution.go
```diff
@@ -1028,3 +1092,96 @@ func (c *Client) SetWorkflowTemplateLabels(namespace, name, prefix string, keyVa
...
+	}
+
+	wfv1Template = wfv1.Template{
+		Name: workflowStepTemplate,
+		Container: &corev1.Container{
+			Image:   "alpine:latest",
+			Command: []string{"sh", "-c"},
+			Args: []string{"apk add curl;" +
+				"curl '" + curlEndpoint + "' -H \"Content-Type: application/json\" -H 'Connection: keep-alive' -H 'Accept: application/json' " +
+				"-H 'Sec-Fetch-Dest: empty' -H 'Authorization: Bearer " + token + "' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.163 Safari/537.36' -H 'Sec-Fetch-Site: same-site' -H 'Sec-Fetch-Mode: cors' -H 'Accept-Language: en-US,en;q=0.9' " +
```
rushtehrani, line 1181: Do we need all these headers?

pkg/workflow_execution.go
```diff
@@ -1028,3 +1092,96 @@ func (c *Client) SetWorkflowTemplateLabels(namespace, name, prefix string, keyVa
...
+
+	token, err := GetBearerToken(namespace)
+	if err != nil {
+		return "", "", "", err, wfv1.Template{}
+	}
+
+	wfv1Template = wfv1.Template{
+		Name: workflowStepTemplate,
+		Container: &corev1.Container{
+			Image:   "alpine:latest",
```
rushtehrani, line 1177: Can't we use [the official curl image](https://hub.docker.com/r/curlimages/curl) so we don't have to do `apk add curl`?

pkg/workflow_execution.go
aleksandrmelnikov: Probably not. I originally took it from one of the web requests from core-ui.

pkg/workflow_execution.go
aleksandrmelnikov: Hmm, let me see

pkg/workflow_execution.go
aleksandrmelnikov: Let me try reducing the number of headers and see if anything breaks.

pkg/workflow_execution.go
aleksandrmelnikov: Good point, taking a look.

pkg/workflow_execution.go
aleksandrmelnikov: Cleaned up the headers. feat/dev-core-144 (merging branch)

pkg/workflow_execution.go
aleksandrmelnikov: Was able to clean this up. feat/dev-core-144 (merging branch)

pkg/workflow_execution.go
aleksandrmelnikov: Changed in feat/dev-core-144

PR #174, opened 2020-05-06
feat: workspaces

rushtehrani:


server/workspace_server.go
```diff
@@ -62,7 +62,7 @@ func (s *WorkspaceServer) CreateWorkspace(ctx context.Context, req *api.CreateWo
 
 func (s *WorkspaceServer) UpdateWorkspaceStatus(ctx context.Context, req *api.UpdateWorkspaceStatusRequest) (*empty.Empty, error) {
 	client := ctx.Value("kubeClient").(*v1.Client)
-	allowed, err := auth.IsAuthorized(client, req.Namespace, "update", "apps/v1", "statefulsets", "")
+	allowed, err := auth.IsAuthorized(client, req.Namespace, "update", "apps", "statefulsets", "")
```
rushtehrani, line 65: @aleksandrmelnikov this needed to be changed to `apps` so that the existing permissions worked. I believe you'll run into a similar issue.

aleksandrmelnikov: I made this change and it appears to be working.

Issue #124, opened 2020-03-24
Add `EnvoyFilter` for authenticating workspaces


rushtehrani:
Example `EnvoyFilter` derived from tests with a different auth API (untested).

There should be an API endpoint exposed that uses [IsAuthorized](https://github.com/onepanelio/core/blob/f110adfd921a652bb6d386442289b37b474c4cb6/server/auth/auth.go#L64) to check if `"create"` is allowed for `StatefulSet` resources.

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: istio-ingress-gateway-auth
  namespace: istio-system
spec:
  filters:
  - listenerMatch:
      portNumber: 443
      listenerType: GATEWAY
    filterName: envoy.lua
    filterType: HTTP
    insertPosition:
      index: FIRST
    filterConfig:
      inlineCode: |
        function get_value(keyvaluepairs, key)
          for value in string.gmatch(keyvaluepairs, key .. "=([^;]*)") do
            return value
          end
        end

        -- Called on the request path.
        function envoy_on_request(request_handle)
          print("Start call to auth.")

          local id_token = nil
          local cookies = request_handle:headers():get("cookie")
          if cookies ~= nil then
            id_token = get_value(cookies, "auth-token")
          end

          local headers =  {
            [":method"] = "HEAD",
            [":path"] = "/apis/v1beta1/auth/token",
            [":authority"] = "onepanel-core.onepanel.svc.local",
            ["x-original-method"] = request_handle:headers():get(":method"),
            ["x-original-authority"] = request_handle:headers():get(":authority"),
            ["x-original-uri"] = request_handle:headers():get(":path")
          }

          if id_token ~= nil then
            headers["authorization"] = "Bearer " .. id_token
          end

          local response_headers, response_body = request_handle:httpCall(
              "outbound_.80_._.onepanel-core.onepanel.svc.cluster.local",
              headers,
              "",
              5000
          )

          if tonumber(response_headers[":status"]) ~= 200 then
            print("Not authorized.")
            request_handle:respond(
              {[":status"] = response_headers[":status"]},
              response_body
            )
          end
        end
```

aleksandrmelnikov:
[WIP]
To Test
- The following branches are used
https://github.com/onepanelio/manifests/pull/54
https://github.com/onepanelio/core-ui/tree/feature/api.updates
https://github.com/onepanelio/core/pull/176

You should spin up an Azure cluster.
[...]

PR #176, opened 2020-05-06
Feat/core.124 add.envoy.filter

aleksandrmelnikov:
closes https://github.com/onepanelio/core/issues/124


server/auth_server.go
```diff
@@ -39,11 +39,13 @@ func (a *AuthServer) IsWorkspaceAuthenticated(ctx context.Context, request *api.
 	}
 	workspaceAndNamespace := xOriginalAuth[0:pos]
 	pieces := strings.Split(workspaceAndNamespace, "--")
-	client := ctx.Value("kubeClient").(*v1.Client)
 	allowed, err := auth.IsAuthorized(client, pieces[1], "create", "apps/v1", "statefulsets", pieces[0])
-	if err != nil || !allowed {
+	if err != nil {
 		return &empty.Empty{}, err
 	}
+	if !allowed {
```
rushtehrani, line 46: No need for this now since this is addressed in this [PR](https://github.com/onepanelio/core/pull/174/files#diff-b49d465c8dbf2c888fe2e2bd0c18e167R81)

server/auth_server.go
```diff
@@ -39,11 +39,13 @@ func (a *AuthServer) IsWorkspaceAuthenticated(ctx context.Context, request *api.
 	}
 	workspaceAndNamespace := xOriginalAuth[0:pos]
 	pieces := strings.Split(workspaceAndNamespace, "--")
-	client := ctx.Value("kubeClient").(*v1.Client)
 	allowed, err := auth.IsAuthorized(client, pieces[1], "create", "apps/v1", "statefulsets", pieces[0])
```
rushtehrani, line 42: I believe we should be using `apps` for group here per our earlier discussion, please verify but making [this change](https://github.com/onepanelio/core/pull/174/files#diff-f9803e0eb312f8ff221f2cfd480173a3R65) worked for me.

server/auth_server.go
aleksandrmelnikov: Made change.

server/auth_server.go
aleksandrmelnikov: This appears to be working on my end as well.

PR #183, opened 2020-05-07
update: simplified request for creating a workflow execution.

Vafilor:


api/workflow.proto
```diff
@@ -92,12 +92,19 @@ service WorkflowService {
...
+    int64 workflowTemplateVersion = 2;
 
+    repeated Parameter parameters = 3;
+    repeated KeyValue labels = 4;
 }
 
 message CreateWorkflowExecutionRequest {    
     string namespace = 1;
-    WorkflowExecution workflowExecution = 2;
+    CreateWorkflowExecutionBody workflowExecution = 2;
```
rushtehrani, line 107: Should we flatten this even more and pull in the `CreateWorkflowExecutionBody` directly in this message?

PR #186, opened 2020-05-08
Refactoring migrations.

aleksandrmelnikov:
- uid is now 63 chars long, to fit k8s requirement
- Updating unique indices to use uid and namespace


pkg/workflow_execution.go
```diff
@@ -379,13 +379,16 @@ func (c *Client) CreateWorkflowExecution(namespace string, workflow *WorkflowExe
 	}
 
 	appendSysUid := true
-	for key, _ := range *opts.Labels {
-		if key == "sys-uid" {
+	for _, param := range opts.Parameters {
+		if param.Name == "sys-uid" {
 			appendSysUid = false
 		}
```
rushtehrani, line 385: Add a `break` here since there is no need to continue the loop once the name if found.

pkg/util/uid/uid.go
```diff
@@ -0,0 +1,16 @@
...
+	"errors"
+	"regexp"
+	"strings"
+)
+
+func GenerateK8sValidName(input string) (string, error) {
+	re, _ := regexp.Compile(`[^a-zA-Z0-9-]{1,}`)
+	cleanUp := strings.ToLower(re.ReplaceAllString(input, `-`))
+	if len(cleanUp) > 30 {
+		return "", errors.New("Length of string exceeds 63, invalid for k8s.")
```
rushtehrani, line 13: 63 => 30

aleksandrmelnikov: What about commenting to explain we're doing this for the k8s sub-domain limitation?

rushtehrani: It's actually more of a limitation set by RFC1123 for subdomains in general. This is specific to us really since we combine uid and namespace in the subdomain.

pkg/util/uid/uid.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strings"
 )
 
-func GenerateK8sValidName(input string) (string, error) {
+func GenerateUid(input string) (string, error) {
```
rushtehrani, line 9: Uid => `UID` per [Go convention](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

pkg/util/uid/uid.go
aleksandrmelnikov: Gotcha, okay, will remove.

PR #191, opened 2020-05-09
Feat/simplify.create.workflow.execution

Vafilor:


api/workspace.proto
```diff
@@ -57,6 +63,12 @@ message Workspace {
 	WorkspaceTemplate workspaceTemplate = 6;
 	WorkspaceStatus status = 7;
 	repeated KeyValue labels = 8;
+
+	string phase = 9;
```
rushtehrani, line 67: Is there a reason these also here? They’re in `WorkspaceStatus`. If this is to simplify, then we should remove `WorkspaceStatus` as a field.

pkg/workspace.go
```diff
@@ -141,17 +162,29 @@ func (c *Client) GetWorkspace(namespace, uid string) (workspace *Workspace, err
...
 		err = nil
 		workspace = nil
+
+		return
 	}
-	if workspace != nil {
-		if err = json.Unmarshal(workspace.ParametersBytes, &workspace.Parameters); err != nil {
-			return
-		}
+	if err != nil {
```
rushtehrani, line 173: This change won’t work because both `err` and `workspace` can be `nil` when a workspace is not found

rushtehrani: Ah, I missed the return above.

api/workspace.proto
```diff
@@ -66,9 +78,23 @@ message WorkspaceStatus {
...
+
+	repeated Parameter parameters = 4;
+	repeated KeyValue labels = 5;
+}
+
 message CreateWorkspaceRequest {
 	string namespace = 1;
-	Workspace workspace = 2;
+
+	CreateWorkspaceBody body = 2;
```
rushtehrani, line 92: I think this should be `workspace` instead of `body`. In gRPC the request looks like this and calling the field `workspace` is more explicit:
image

Thoughts?

rushtehrani: On the other hand, looking at generated SDK, `body` makes sense in that context. I'm still testing but going to resolve this conversation.

api/workspace.proto
Vafilor: Removed it, having it in status is good.

PR #195, opened 2020-05-11
feature: added path for workspaces

Vafilor:


pkg/workspace_types.go
```diff
@@ -42,6 +42,7 @@ type Workspace struct {
 	WorkspaceTemplate        *WorkspaceTemplate `db:"workspace_template" valid:"-"`
 	WorkspaceTemplateID      uint64             `db:"workspace_template_id"`
 	WorkspaceTemplateVersion uint64             `db:"workspace_template_version"`
+	Path                     string             `db:"path"` // the path to the workspace, a url that you can access via http
```
rushtehrani, line 45: should this be called `URL` instead of `Path`?

Issue #208, opened 2020-05-18
Expose IsAuthorized as a service


rushtehrani:
We need to expose [IsAuthorized](https://github.com/onepanelio/core/blob/master/server/auth/auth.go#L64) as a service so that it can be used by UI to check access and toggle display of actions on Workspaces, Workspace Templates, Workflows, Workflow Templates, CronWorkflows

aleksandrmelnikov:
I tested this endpoint with `auth token` from cli, and then the default-token in my namespace, alex.
- auth token worked
- alex token did not

But the logic should be checked.

PR #210, opened 2020-05-18
Feat/core.208 expose.isauthorized

aleksandrmelnikov:
closes https://github.com/onepanelio/core/issues/208


server/auth_server.go
```diff
@@ -46,40 +46,71 @@ func (a *AuthServer) IsWorkspaceAuthenticated(ctx context.Context, request *api.
...
+	if err != nil {
+		return
+	}
+	res = &api.IsValidTokenResponse{}
+	res.Domain = config["ONEPANEL_DOMAIN"]
+
+	return res, nil
+}
+
+func (a *AuthServer) isAuthorized(err error, client *v1.Client) (*api.IsValidTokenResponse, error, bool) {
```
rushtehrani, line 94: Bit confused about this, is this combining token validation and authorization?

aleksandrmelnikov: Yes.
My understanding is that the token validation is how we know the user is valid.
The authorization is to see if the user has resource access.

server/auth_server.go
```diff
@@ -46,40 +46,71 @@ func (a *AuthServer) IsWorkspaceAuthenticated(ctx context.Context, request *api.
...
 
+func (a *AuthServer) IsAuthorized(ctx context.Context, request *api.IsAuthorizedRequest) (res *api.IsAuthorizedResponse, err error) {
+	res = &api.IsAuthorizedResponse{}
+	if ctx == nil {
+		res.Authorized = false
+		return res, status.Error(codes.Unauthenticated, "Unauthenticated.")
+	}
+	//User auth check
+	client := ctx.Value("kubeClient").(*v1.Client)
+	_, err2, _ := a.isAuthorized(err, client)
```
rushtehrani, line 57: why `err2`?

aleksandrmelnikov: Err, that was automatically generated.

aleksandrmelnikov: Just renamed.

server/auth_server.go
rushtehrani: Should this be two separate methods instead of one? Or is this intent of this method is for really to check the validity of the token? If so, `isValidToken` might be a better name?

server/auth_server.go
```diff
@@ -46,40 +46,71 @@ func (a *AuthServer) IsWorkspaceAuthenticated(ctx context.Context, request *api.
...
+	if err != nil {
+		return
+	}
+	res = &api.IsValidTokenResponse{}
+	res.Domain = config["ONEPANEL_DOMAIN"]
+
+	return res, nil
+}
+
+func (a *AuthServer) isAuthorized(err error, client *v1.Client) (*api.IsValidTokenResponse, error, bool) {
```
rushtehrani, line 94: If only `err` is being used, maybe this method should only return that instead of three values?

aleksandrmelnikov: Agreed, refactored.

server/auth_server.go
aleksandrmelnikov: The main intent is to validate the user.
They are a valid user, so they can make this request.

I combined the two calls there because that's how it seems we're figuring out the user.

I do agree with the name change, so I made it.

rushtehrani:
Changes look good, will test functionality shortly.

PR #211, opened 2020-05-18
feat: inject default artifact repository for input/output artifacts without creds defined

### LINKED ISSUE #207: Inject S3 system config info for `inputs`
When user defines the following:
```
inputs:
      artifacts:
        - name: data
          path: /mnt/data/datasets/
          s3:
            key: '{{workflow.parameters.dataset-path}}'
```

Automatically inject the remaining fields:
```
            accessKeySecret:
              key: artifactRepositoryS3AccessKey
              name: onepanel
            bucket: <bucket-from-system-config>
            endpoint: s3.amazonaws.com
            insecure: <value-from-system-config>
            secretKeySecret:
              key: artifactRepositoryS3SecretKey
              name: onepanel
```

This is a workaround until [this issue](https://github.com/argoproj/argo/pull/3030) is addressed.

rushtehrani:
closes #207

### DIFF
--- a/pkg/workflow_execution.go
+++ b/pkg/workflow_execution.go
@@ -85,23 +85,6 @@ func UnmarshalWorkflows(wfBytes []byte, strict bool) (wfs []wfv1.Workflow, err e
 	return
 }
 
-func addEnvToTemplate(template *wfv1.Template, key string, value string) {
-	//Flag to prevent over-writing user's envs
-	overwriteUserEnv := true
-	for _, templateEnv := range template.Container.Env {
-		if templateEnv.Name == key {
-			overwriteUserEnv = false
-			break
-		}
-	}
-	if overwriteUserEnv {
-		template.Container.Env = append(template.Container.Env, corev1.EnvVar{
-			Name:  key,
-			Value: value,
-		})
-	}
-}
-
 func appendArtifactRepositoryConfigIfMissing(artifact *wfv1.Artifact, namespaceConfig *NamespaceConfig) bool {
 	if artifact.S3 != nil && artifact.S3.Key != "" && artifact.S3.Bucket == "" {
 		s3Config := namespaceConfig.ArtifactRepository.S3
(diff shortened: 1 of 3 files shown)


pkg/workflow_execution.go
```diff
@@ -155,18 +162,30 @@ func (c *Client) injectAutomatedFields(namespace string, wf *wfv1.Workflow, opts
 			},
 		})
 
+		// Extend artifact credentials if only key is provided
+		if len(template.Outputs.Artifacts) > 0 {
+			for j, artifact := range template.Outputs.Artifacts {
```
Vafilor, line 167: We can omit the `if len` check here, loop won't do anything if there are no items.

pkg/workflow_execution.go
```diff
@@ -155,18 +162,30 @@ func (c *Client) injectAutomatedFields(namespace string, wf *wfv1.Workflow, opts
...
+				if appendArtifactRepositoryConfigIfMissing(&artifact, namespaceConfig) {
+					wf.Spec.Templates[i].Outputs.Artifacts[j] = artifact
+				}
+			}
+		}
+
+		if len(template.Inputs.Artifacts) > 0 {
```
Vafilor, line 174: We can omit the `if len` check here, loop won't do anything if there are no items.

pkg/workflow_execution.go
```diff
@@ -85,21 +85,20 @@ func UnmarshalWorkflows(wfBytes []byte, strict bool) (wfs []wfv1.Workflow, err e
...
-			overwriteUserEnv = false
-			break
-		}
-	}
-	if overwriteUserEnv {
-		template.Container.Env = append(template.Container.Env, corev1.EnvVar{
-			Name:  key,
-			Value: value,
-		})
+func appendArtifactRepositoryConfigIfMissing(artifact *wfv1.Artifact, namespaceConfig *NamespaceConfig) bool {
```
Vafilor, line 88: Can we add a quick blurb here that it returns true if config was missing, false otherwise? (I think that's the return)

rushtehrani:
@Vafilor made code updates to https://github.com/onepanelio/core/pull/213 based on your comments. I think we can close this PR and use that since that contains these changes as well.

Issue #171, opened 2020-05-04
When archiving a workflow template, ensure no orphan objects remain.


rushtehrani:
- Clean up workflow template in k8s
- Clean up any cron workflows in k8s
- Clean up any workflow executions in k8s

Note that changes in https://github.com/onepanelio/core/pull/212 will only work if the `WorkflowTemplate` resource is deleted, otherwise the user gets a "Workflow template already exists" error.

rushtehrani:
Noting this for testing: `ArchiveWorkflowTemplate` also gets called in case `WorkspaceTemplate` creation fails:
https://github.com/onepanelio/core/blob/master/pkg/workspace_template.go#L445
https://github.com/onepanelio/core/blob/master/pkg/workspace_template.go#L459
https://github.com/onepanelio/core/blob/master/pkg/workspace_template.go#L464

rushtehrani:
It looks like `workflow_template_versions` are not archived. Below is an example of a workflow template that was recreated using the same name as an archived one and the versions of the previous template are still showing up.
![image](https://user-images.githubusercontent.com/1211823/82720317-559ae300-9c67-11ea-83ae-70975148d049.png)

PR #226, opened 2020-05-20
Feat/core.171 archiving.workflow.template.clean.up

aleksandrmelnikov:
closes #https://github.com/onepanelio/core/issues/171


pkg/cron_workflow.go
```diff
@@ -668,3 +650,50 @@ func cronWorkflowColumns(extraColumns ...string) []string {
...
+	cronWorkflow := &CronWorkflow{}
+	if err = c.DB.Get(cronWorkflow, query, args...); err != nil {
+		return err, nil
+	}
+
+	return nil, cronWorkflow
+}
+
+func (c *Client) DeleteCronWorkflowDb(namespace, uid string) error {
+	query := `DELETE FROM cron_workflows
```
rushtehrani, line 678: Can this query be rewritten using Squirrel?

pkg/workflow_execution.go
```diff
@@ -1463,3 +1463,24 @@ func (c *Client) getWorkflowExecutionAndTemplate(namespace string, uid string) (
...
+		if strings.Contains(err.Error(), "not found") {
+			return nil
+		}
+		return err
+	}
+	return nil
+}
+
+func (c *Client) DeleteWorkflowExecutionDb(workflowUid string) error {
+	query := `DELETE FROM workflow_executions
```
rushtehrani, line 1479: Can this query be rewritten using Squirrel?

pkg/cron_workflow.go
Vafilor: Not with `using`.

But we can do a GET to load the cron workflow and then delete by id.

rushtehrani: Got it, for some reason I thought Squirrel allowed custom SQL to be mixed with their methods.

rushtehrani: If there isn't a cleaner way to do the `where` values, then this is fine.

pkg/workflow_execution.go
aleksandrmelnikov: Refactored

pkg/cron_workflow.go
aleksandrmelnikov: Added a refactoring. The only thing I am unsure of is I'm not using namespace anymore. Just the uid of the cron workflow.
It worked when I tested locally, but I didn't test multiple cron workflows.

rushtehrani: I'd use the namespace too to be safe. We should use that in all queries.

aleksandrmelnikov: Cron Workflows do not have a namespace column in the db.
The Workflow Template that gets selected does.

I'm testing with multiple cron workflows and seeing some constraint violation issues, investigating.

rushtehrani: Ah, they definitely need to have `namespace`. Need to fix that.

pkg/cron_workflow.go
```diff
@@ -660,11 +645,75 @@ func (c *Client) GetCronWorkflowStatisticsForTemplates(workflowTemplates ...*Wor
 }
 
 func cronWorkflowColumns(extraColumns ...string) []string {
```
Vafilor, line 647: Let's call this `getCronWorkflowColumns` so it matches the other, similarly named methods.

Here's one of the other ones,

```
func getWorkspaceStatusColumns(alias string, destination string, extraColumns ...string) []string {
	columns := []string{"phase", "started_at", "paused_at", "terminated_at"}
	return formatColumnSelect(columns, alias, destination, extraColumns...)
}
```

pkg/cron_workflow.go
```diff
@@ -660,11 +645,75 @@ func (c *Client) GetCronWorkflowStatisticsForTemplates(workflowTemplates ...*Wor
...
+	results := []string{"cw.id", "cw.created_at", "cw.uid", "cw.name", "cw.workflow_template_version_id", "cw.manifest", "cw.namespace"}
 
 	for _, str := range extraColumns {
 		results = append(results, str)
 	}
 
 	return results
 }
+
+func (c *Client) SelectCronWorkflowWithExtraColumns(namespace, uid string, extraColumns ...string) (error, *CronWorkflow) {
```
Vafilor, line 657: Let's change the return type order here.
```
(*CronWorkflow, error)
```

pkg/cron_workflow.go
```diff
@@ -660,11 +645,75 @@ func (c *Client) GetCronWorkflowStatisticsForTemplates(workflowTemplates ...*Wor
...
+	cronWorkflow := &CronWorkflow{}
+	if err = c.DB.Get(cronWorkflow, query, args...); err != nil {
+		return err, nil
+	}
+
+	return nil, cronWorkflow
+}
+
+func (c *Client) DeleteCronWorkflowDb(namespace, uid string) error {
+	query, args, err := sb.Select("id").From("cron_workflows").
```
Vafilor, line 681: Since the namespace and uid are now both on cron workflows, let's just use that to delete them. No need to get the id first.

pkg/cron_workflow.go
rushtehrani: Should this also be in `cron_workflow_types.go` like the other methods?

pkg/cron_workflow.go
```diff
@@ -660,11 +645,75 @@ func (c *Client) GetCronWorkflowStatisticsForTemplates(workflowTemplates ...*Wor
...
+		return err
+	}
+
+	if _, err := c.DB.Exec(query, args...); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Client) DeleteCronWorkflow(namespace, uid string) error {
```
Vafilor, line 710: What do you think about renaming this to `DeleteArgoCronWorkflow`?

pkg/workflow_template.go
```diff
@@ -608,7 +612,7 @@ func (c *Client) ListWorkflowTemplates(namespace string, paginator *pagination.P
 }
 
 func (c *Client) ArchiveWorkflowTemplate(namespace, uid string) (archived bool, err error) {
-	workflowTemplate, err := c.getWorkflowTemplate(namespace, uid, 0)
+	workflowTemplate, err := c.getWorkflowTemplate(namespace, uid, 0) //version=0 means latest
```
Vafilor, line 615: Maybe we should add a `getLatestWorkflowTemplate` method so we don't need to pass in a weird 0 for version. We'd still have that method, but add this new convenience one. 

What do you guys think?
@aleksandrmelnikov @rushtehrani

aleksandrmelnikov: Generally, I like the idea. Have to use go docs to tell the user to use it ;).

pkg/cron_workflow.go
rushtehrani: This also shouldn't be an exported method.

pkg/workflow_execution.go
```diff
@@ -1463,3 +1463,28 @@ func (c *Client) getWorkflowExecutionAndTemplate(namespace string, uid string) (
...
+	if err != nil {
+		if strings.Contains(err.Error(), "not found") {
+			return nil
+		}
+		return err
+	}
+	return nil
+}
+
+func (c *Client) DeleteWorkflowExecutionDb(workflowUid string) error {
```
Vafilor, line 1478: I believe this needs a namespace to be uniquely identified?

Vafilor: Also, I think this should be archiving it and not actually deleting it? Might be wrong.

db/20200521145827_add_is_archived_to_cron_wfs_and_wf_execs.sql
```diff
@@ -0,0 +1,8 @@
+-- +goose Up
+-- SQL in this section is executed when the migration is applied.
+ALTER TABLE cron_workflows ADD COLUMN is_archived BOOL;
+ALTER TABLE workflow_executions ADD COLUMN is_archived BOOL;
```
Vafilor, line 4: @aleksandrmelnikov Let's add a `DEFAULT false;` to these columns. I forget if that will make existing records have `false` for the value. If they don't, we need to run an update query so existing ones are marked as archived false.

@rushtehrani Should these columns be nullable? I'm leaning towards no, though it might be redundant with a DEFAULT set.

rushtehrani: They should not be nullable. Only acceptable values are `true` or `false` in this case.

Vafilor: @rushtehrani Makes sense. Okay, we'll have to fix the other tables too, I believe they are nullable (workflow and workspace templates).

Sample query for updating an existing column:
```sql
ALTER TABLE workflow_templates ADD is_archived BOOL;
UPDATE workflow_templates SET is_archived = false;
ALTER TABLE workflow_templates ALTER COLUMN is_archived SET NOT NULL;
```

pkg/workflow_execution.go
```diff
@@ -1463,3 +1463,28 @@ func (c *Client) getWorkflowExecutionAndTemplate(namespace string, uid string) (
...
+	query, args, err := sb.Delete("workflow_executions").Where(sq.Eq{
+		"uid": workflowUid,
+	}).ToSql()
+	if err != nil {
+		return err
+	}
+
+	if _, err := c.DB.Exec(query, args); err != nil {
```
Vafilor, line 1486: This should be
```go
c.DB.Exec(query, args...)
```

It trips me up all the time >.<

rushtehrani: This should be:

```
err := sb.Delete("workflow_executions").Where(sq.Eq{
	"uid": workflowUid,
         "namespace": namespace,
  }).RunWith(c.DB).Exec()
```
Since we are not marshalling into a struct. Also note the namespace in `Where` so we need to pass that here.

Another note, I don't see a point to having a separate DB method here. This is a small and query can easily be placed in the calling method. I don't see a case where we will delete only database records and not the corresponding K8S resource.

db/20200521145827_add_is_archived_to_cron_wfs_and_wf_execs.sql
aleksandrmelnikov: Refactored to this:
```
ALTER TABLE cron_workflows ADD COLUMN is_archived BOOL DEFAULT false NOT NULL;
ALTER TABLE workflow_executions ADD COLUMN is_archived BOOL DEFAULT false NOT NULL;
```
This creates default false, and makes it not null

pkg/workflow_execution.go
aleksandrmelnikov: You are right about namespace.

Regarding archive vs delete, I added a separate method to archive.

I ended up removing this function because it's not being used anywhere.
- Plus, the sql inside doesn't save much code (per our discussion this morning); so it shouldn't be a function anyway.

pkg/workflow_execution.go
aleksandrmelnikov: Refactored per feedback.

pkg/cron_workflow.go
aleksandrmelnikov: Refactored

pkg/cron_workflow.go
aleksandrmelnikov: This branch should have the namespace added and working, along with refactoring.

pkg/cron_workflow.go
aleksandrmelnikov: @Vafilor Question about this one, it's being used by workflow_template.go?
So making it un-exported will cause an issue right?

pkg/cron_workflow.go
aleksandrmelnikov: I think this is no longer relevant because the refactoring removed this method?

pkg/cron_workflow.go
aleksandrmelnikov: I think this is outdated, method is gone

pkg/workflow_template.go
aleksandrmelnikov: Refactored

pkg/cron_workflow.go
rushtehrani: @aleksandrmelnikov no, that's in the same package.

pkg/cron_workflow.go
```diff
@@ -660,11 +658,35 @@ func (c *Client) GetCronWorkflowStatisticsForTemplates(workflowTemplates ...*Wor
...
+	results := []string{"cw.id", "cw.created_at", "cw.uid", "cw.name", "cw.workflow_template_version_id", "cw.manifest", "cw.namespace"}
 
 	for _, str := range extraColumns {
 		results = append(results, str)
 	}
 
 	return results
 }
+
+func (c *Client) SelectCronWorkflowWithWorkflowTemplateVersion(namespace, uid string, extraColumns ...string) (error, *CronWorkflow) {
```
rushtehrani, line 670: Does this need to be public?

pkg/cron_workflow.go
```diff
@@ -570,14 +566,15 @@ func (c *Client) cronWorkflowSelectBuilder(namespace string, workflowTemplateUid
 }
 
 func (c *Client) cronWorkflowSelectBuilderNamespaceName(namespace string, uid string) sq.SelectBuilder {
-	sb := sb.Select("cw.id", "cw.created_at", "cw.uid", "cw.name", "cw.workflow_template_version_id").
+	sb := sb.Select("cw.id", "cw.created_at", "cw.uid", "cw.name", "cw.workflow_template_version_id", "cw.namespace").
```
rushtehrani, line 569: should use method here to get the columns as mentioned in [this comment](https://github.com/onepanelio/core/pull/226/files#r428964039)

pkg/cron_workflow.go
aleksandrmelnikov: My bad, I thought it was upper case for some reason. Refactored.

pkg/cron_workflow.go
aleksandrmelnikov: Good point, refactored

pkg/cron_workflow.go
aleksandrmelnikov: Makes sense. The mentioned method is equal to what this one does, so I removed it and used that one.

PR #242, opened 2020-05-23
increase column size to 63 for cron workflows and workflow executions

rushtehrani:


db/20200522164526_alter_workflow_executions_column_sizes.sql
```diff
@@ -0,0 +1,7 @@
+-- +goose Up
+ALTER TABLE workflow_executions ALTER COLUMN uid TYPE varchar(63);
+ALTER TABLE workflow_executions ALTER COLUMN name TYPE varchar(63);
+
+-- +goose Down
+ALTER TABLE workflow_executions ALTER COLUMN uid TYPE varchar(30);
+ALTER TABLE workflow_executions ALTER COLUMN uid TYPE text;
```
Vafilor, line 7: Think one of these needs to be a name @rushtehrani

rushtehrani: Yes, good catch!

PR #260, opened 2020-05-26
feature: added migration with jupyterlab

Vafilor:


db/20200525160514_add_jupyter_workspace_template.go
```diff
@@ -0,0 +1,137 @@
...
+	return client, nil
+}
+
+func Up20200525160514(tx *sql.Tx) error {
+	client, err := getClient()
+	if err != nil {
+		return err
+	}
+
+	namespaces, err := client.ListOnepanelEnabledNamespaces()
```
rushtehrani, line 100: Can grabbing the default namespace be extracted into another file like `migration.go` so we can reuse it for other Go migrations?

db/20200525160514_add_jupyter_workspace_template.go
```diff
@@ -0,0 +1,137 @@
...
+       args:
+       - SLACK_USERNAME=onepanel SLACK_TITLE="Your workspace is ready" SLACK_ICON=https://www.gravatar.com/avatar/5c4478592fe00878f62f0027be59c1bd SLACK_MESSAGE="Your workspace is now running" ./slack-notify
+       command:
+       - sh
+       - -c`
+
+const jupyterLabTemplateName = "jupyterlab"
+
+func init() {
+	goose.AddMigration(Up20200525160514, Down20200525160514)
```
rushtehrani, line 73: Can this also be put in a common file like `migration.go` so that we can add more migrations in one place?

db/20200525160514_add_jupyter_workspace_template.go
rushtehrani: Just realized this is a single method call so there is no point in putting it in another method.

Issue #264, opened 2020-05-26
Workspace Template `postExecutionWorkflow` should be optional


Vafilor:
Currently Workspace Templates crash on this line
https://github.com/onepanelio/core/blob/develop/pkg/workspace_template.go#L346

Trying to access `spec.PostExecutionWorkflow.Entrypoint`

PR #284, opened 2020-05-28
fix: added a default post execution workflow if there wasn't one alre…

Vafilor:
…ady.

closes onepanelio/core#264

### DIFF
--- a/pkg/workspace_template.go
+++ b/pkg/workspace_template.go
@@ -259,6 +259,18 @@ kind: PersistentVolumeClaim
 metadata:
   name: {{inputs.parameters.sys-pvc-name}}-{{workflow.parameters.sys-uid}}-0
 `
+
+	if spec.PostExecutionWorkflow == nil {
+		spec.PostExecutionWorkflow = &wfv1.WorkflowTemplateSpec{}
+		spec.PostExecutionWorkflow.Entrypoint = "main"
+		spec.PostExecutionWorkflow.Templates = []wfv1.Template{
+			{
+				Name: "main",
+				DAG:  &wfv1.DAGTemplate{},
+			},
+		}
+	}
+
 	templates := []wfv1.Template{
 		{
 			Name: "workspace",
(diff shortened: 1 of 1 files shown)


pkg/workspace_template.go
```diff
@@ -259,6 +259,18 @@ kind: PersistentVolumeClaim
 metadata:
   name: {{inputs.parameters.sys-pvc-name}}-{{workflow.parameters.sys-uid}}-0
 `
+
+	if spec.PostExecutionWorkflow == nil {
```
rushtehrani, line 263: Is this going to display in the UI? I think we should check if this is `nil` and bypass it instead of adding an empty DAG.

Vafilor: @rushtehrani I think I get it now.

This was originally put in here to make sure this line worked.

https://github.com/onepanelio/core/blob/fix/onepanelio.core.264-post.execution.workflow.optional/pkg/workspace_template.go#L358

But I think we can instead check if it is not nil, and if so, then we can append it.

rushtehrani: Correct, it looks like it’s getting checked [here](https://github.com/onepanelio/core/blob/ca1c672f0d0cd1f1f8f55e69442509c1ddd46f6f/pkg/workspace_template.go#L425) so it won’t attempt to append the template.

Vafilor: @rushtehrani Correct, but it does fail [here](https://github.com/onepanelio/core/blob/ca1c672f0d0cd1f1f8f55e69442509c1ddd46f6f/pkg/workspace_template.go#L358) but I think we can move that block into the one you indicated.

rushtehrani: Sorry, I was just saying it’s already getting checked in that line so you only need to check it on the line you mentioned.

Vafilor: Ahh, gotcha. Okay, let me get an image up to test.

Vafilor: Updated

Issue #282, opened 2020-05-28
Expose artifact path configuration in `params.yaml`


rushtehrani:
We need to expose [artifactRepositoryS3KeyFormat](https://github.com/onepanelio/manifests/blob/939cd1209e352e181b1e9d376bfa3616fc31b2bb/vars/workflow-config-map-hidden.env#L3) in `params.yaml` with the default set:

```
artifactRepository:
  s3:
    keyFormat:  artifacts/{{workflow.namespace}}/{{workflow.name}}/{{pod.name}}
```

This is used in these locations in manifests (don't believe there are any changes needed to these files):
- [onepanel/base/configmap-onepanel-defaultnamespace.yaml](https://github.com/onepanelio/manifests/blob/master/common/onepanel/base/configmap-onepanel-defaultnamespace.yaml#L10)
- [argo/base/config-map.yaml](https://github.com/onepanelio/manifests/blob/939cd1209e352e181b1e9d376bfa3616fc31b2bb/common/argo/base/config-map.yaml#L15)

This however is hardcoded in `core` and needs to be updated there:
- https://github.com/onepanelio/core/blob/afb7af1bbe7ea4f47c26a3615d8a958375483c32/pkg/workflow_execution.go#L828
- https://github.com/onepanelio/core/blob/afb7af1bbe7ea4f47c26a3615d8a958375483c32/pkg/workflow_execution.go#L908

PR #290, opened 2020-05-29
update: hard-coded artifacts key format with the new config.

Vafilor:
closes onepanelio/core#282 (along with manifests PR)

### DIFF
--- a/pkg/workflow_execution.go
+++ b/pkg/workflow_execution.go
@@ -824,7 +824,9 @@ func (c *Client) GetWorkflowExecutionLogs(namespace, uid, podName, containerName
 			return nil, util.NewUserError(codes.InvalidArgument, "Invaild range.")
 		}
 		opts.SetRange(0, int64(endOffset))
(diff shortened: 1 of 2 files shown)


pkg/workflow_execution.go
```diff
@@ -824,8 +824,7 @@ func (c *Client) GetWorkflowExecutionLogs(namespace, uid, podName, containerName
 			return nil, util.NewUserError(codes.InvalidArgument, "Invaild range.")
 		}
 		opts.SetRange(0, int64(endOffset))
-
-		stream, err = s3Client.GetObject(config.ArtifactRepository.S3.Bucket, "artifacts/"+namespace+"/"+uid+"/"+podName+"/"+containerName+".log", opts)
+		stream, err = s3Client.GetObject(config.ArtifactRepository.S3.Bucket, config.ArtifactRepository.S3.KeyFormat+"/"+containerName+".log", opts)
```
rushtehrani, line 827: @Vafilor this won't work as the tokens in `config.ArtifactRepository.S3.KeyFormat` need to be replaced with the actual values.

Vafilor: @rushtehrani Updated.

Issue #266, opened 2020-05-26
IsAuthorized API changes


rushtehrani:
We need to combine [IsWorkspaceAuthorized](https://github.com/onepanelio/core/blob/master/server/auth_server.go#L21) and [IsAuthorized](https://github.com/onepanelio/core/blob/master/server/auth_server.go#L49) to allow for the following scenarios:
- Workspaces will check authorization by posting to same endpoint. They use cookies so this combined call will support those.
- Calls to APIs residing in Workspaces (for example AirSim APIs) will need Bearer token header, so this combined API should support that as well.
- The combined API is also going to be called by Onepanel UI to check whether an action should be displayed. In this case the call is made via a Bearer token in the header, similar to bullet #2.
- Authenticate services in Workflows that will be exposed in UI, for example TensorBoard and Terminal

PR #338, opened 2020-06-05
feat: Centralize authorization to one endpoint

Vafilor:
**What this PR does**:

Removes the authorization endpoint for workspaces and puts that logic into the single IsAuthorized endpoint.

Fixes onepanelio/core#266

Also see the Manifests PR that removes fqdn from headers.


rushtehrani:
@Vafilor please [address](https://app.codacy.com/gh/onepanelio/core/pullRequest?prid=5650225)

Vafilor:
> @Vafilor please [address](https://app.codacy.com/gh/onepanelio/core/pullRequest?prid=5650225)

@rushtehrani done.

main.go
```diff
@@ -98,7 +103,7 @@ func startRPCServer(db *v1.DB, kubeConfig *v1.Config) {
 	api.RegisterWorkflowServiceServer(s, server.NewWorkflowServer())
 	api.RegisterSecretServiceServer(s, server.NewSecretServer())
 	api.RegisterNamespaceServiceServer(s, server.NewNamespaceServer())
-	api.RegisterAuthServiceServer(s, server.NewAuthServer())
+	api.RegisterAuthServiceServer(s, server.NewAuthServer(namespaces, config))
```
rushtehrani, line 106: @Vafilor I wonder if a better pattern would be to cache `SystemConfig` when `GetSystemConfig()` is called the first time and subsequent calls to `GetSystemConfig()` would use the cached value. `GetSytemConfig()` is called in multiple places in code so this will make things more efficient across the board.

server/auth_server.go
```diff
@@ -2,48 +2,35 @@ package server
...
-	if pos == -1 {
-		return &empty.Empty{}, errors.New("Error parsing x-original-authority. No '.' character.")
+
+// NewAuthServer creates a new AuthServer with some pre-cached data.
+// - namespaces are all of the user namespaces in the system
+// - config is the system config.
+//     ONEPANEL_DOMAIN is required at a minimum.
+func NewAuthServer(namespaces []*v1.Namespace, config map[string]string) *AuthServer {
+	server := &AuthServer{
+		fqdns: make(map[string]string),
```
rushtehrani, line 26: Bit confused about this, why do we have multiple `fqdns` and why are namespaces involved. `fqdn` is the application domain in [params.yaml](https://github.com/onepanelio/manifests/blob/master/common/onepanel/base/vars.yaml#L50)

rushtehrani: Looks like you're setting the workspace names in here. Wouldn't it be better to just pass through auth for `fdqn` from `params.yaml`?

main.go
Vafilor: @rushtehrani Pushed up a change for this. Let me know what you think.

server/auth_server.go
Vafilor: @rushtehrani This is bad naming on my part. 

First - do you mean pass in the fqdn instead of the entire config? I'm up for that.

Second - this is supposed to be the suffix part of  `workspace.name--(suffix)`.

When we check permissions for workspace related items, it's a quick-ish way to get the expected full path of it -` workspace.name--namespace.domain`.

We prepend the `workspace.name`.

main.go
rushtehrani: @Vafilor looks good, you don't need to pass it in anymore, you can just call `GetSystemConfig()` in `NewAuthServer`.

Vafilor: @rushtehrani 

Roger that - this answers a question in a comment below.

pkg/config.go
```diff
@@ -22,24 +22,32 @@ func (c *Client) getConfigMap(namespace, name string) (configMap *ConfigMap, err
...
+// GetSystemConfig loads various system configurations and bundles them into a map.
+// The configuration is cached once it is loaded, and that cached value is used from here on out.
 func (c *Client) GetSystemConfig() (config map[string]string, err error) {
-	namespace := "onepanel"
-	configMap, err := c.getConfigMap(namespace, "onepanel")
-	if err != nil {
-		return
-	}
-	config = configMap.Data
+	if c.systemConfig == nil {
```
rushtehrani, line 28: We need to figure out a better way to cache this for the entire application and not just at request level.

PR #346, opened 2020-06-11
feat: Centralize authorization to one endpoint

Vafilor:
**What this PR does**:

Removes the authorization endpoint for workspaces and puts that logic into the single IsAuthorized endpoint.

Fixes onepanelio/core#266

Also see the Manifests PR that removes fqdn from headers.


pkg/config.go
```diff
@@ -22,23 +26,33 @@ func (c *Client) getConfigMap(namespace, name string) (configMap *ConfigMap, err
...
+// The configuration is cached once it is loaded, and that cached value is used from here on out.
 func (c *Client) GetSystemConfig() (config map[string]string, err error) {
+	if c.systemConfig != nil {
+		return c.systemConfig, nil
+	}
+
 	namespace := "onepanel"
-	configMap, err := c.getConfigMap(namespace, "onepanel")
-	if err != nil {
+	configMap, configMapErr := c.getConfigMap(namespace, "onepanel")
```
rushtehrani, line 37: why not just use `err`?

Vafilor: Originally I got an err shadowed compilation error. 
But I'm not getting it now - pushing up a fix.

Issue #348, opened 2020-06-11
Runtime variables and system config change reload


Vafilor:
- node pool configs and domain are saved in workspace workflow templates so to update them you’d have to manually update each workspace workflow template
- system config to change database should restart onepanel-core pod.
- workspace URLs contain the domain in the database, if the domain is updated, this will cause an issue.

PR #352, opened 2020-06-16
feat: runtime variables and system config change reload

Vafilor:
**What this PR does**:

Updates the workspace templates so that runtime variables like the DOMAIN are injected when the workspace is is created instead of when the workspace template is created.

This also updates the server so that it reloads data upon a detected configuration change, so variables like DOMAIN are always up to date.

Fixes onepanelio/core#348

For the runtime variables, there are 3 main places of interest.
- When we create a template for viewing, as when we generate a dynamic template while creating/editing a workspace template. In this situation, we need to inject runtime variables for things like the Machine Type dropdown.
- Whenever we get a Workspace Template, we need to modify the manifest to include the variables.
- When we create the Workflow Execution for a Workspace. The Argo WorkflowTemplate needs to have the appropriate variables injected into it.

NOTE:
If you attempt to view the YAML in the workflow execution, it will not have the injected variables as it gets the Workflow Execution, and we do not know to inject the runtime variables there. We might want to eventually add a method to get the workflow execution yaml for a Workspace to get around this.

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"net"
 	"net/http"
+	"sync"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -47,6 +48,8 @@ func main() {
 	// We do this when the configuration has been changed, so the server has the latest configuration
 	stopCh := make(chan struct{})
 
+	lock := sync.RWMutex{}
+
 	go func() {
 		kubeConfig := v1.NewConfig()
 		client, err := v1.NewClient(kubeConfig, nil, nil)
@@ -57,6 +60,7 @@ func main() {
 		go watchConfigmapChanges(client, "onepanel", stopCh, func(configMap *corev1.ConfigMap) error {
 			log.Printf("Configmap changed")
 			stopCh <- struct{}{}
+			lock.Unlock()
 
 			return nil
 		})
@@ -78,9 +82,19 @@ func main() {
 
 			s := startRPCServer(db, kubeConfig, sysConfig, stopCh)
 
-			<-stopCh
+			lock.Lock()
+
+			log.Printf("Past first lock")
+
+			lock.Lock()
+
+			//<-stopCh
+			log.Printf("Stop channel is done, stopping server")
 
 			s.Stop()
+
+			log.Printf("Stop channel is done, server stopped")
+			lock.Unlock()
 		}
 	}()
 
--- a/main.go
+++ b/main.go
@@ -5,6 +5,12 @@ import (
 	"flag"
 	"fmt"
 	_ "github.com/onepanelio/core/db"
+	corev1 "k8s.io/api/core/v1"
+	apiv1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	k8runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
 	"net"
 	"net/http"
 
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -103,9 +124,27 @@ func startRPCServer(db *v1.DB, kubeConfig *v1.Config, sysConfig v1.SystemConfig)
...
 
-	if err := s.Serve(lis); err != nil {
-		log.Fatalf("Failed to serve RPC server: %v", err)
+	client, err := v1.NewClient(kubeConfig, db, sysConfig)
+	if err != nil {
+		log.Fatalf("Unable to create client: %v", err)
 	}
+
+	go Run(client, "onepanel", stopCh, func(configMap *corev1.ConfigMap) error {
+		log.Printf("Configmap changed")
```
Vafilor, line 133: We can get rid of this, it's more for debugging.

main.go
```diff
@@ -103,9 +124,27 @@ func startRPCServer(db *v1.DB, kubeConfig *v1.Config, sysConfig v1.SystemConfig)
...
+
+		return nil
+	})
+
+	go func() {
+		if err := s.Serve(lis); err != nil {
+			log.Fatalf("Failed to serve RPC server: %v", err)
+		}
+
+		log.Infof("GRPC finished")
```
Vafilor, line 144: We can get rid of this, it's more for debugging.

pkg/workspace_template.go
```diff
@@ -248,6 +256,23 @@ func createStatefulSetManifest(spec *WorkspaceSpec, config map[string]string) (s
...
+			},
+		},
+	}
+
+	if withRuntimeVars {
+		template.Spec = corev1.PodSpec{
+			NodeSelector: map[string]string{
+				config["applicationNodePoolLabel"]: "{{workflow.parameters.sys-node-pool}}",
+			},
+			Containers: workspaceSpec.Containers,
```
rushtehrani, line 272: This needs to be updated to `spec`

rushtehrani:
@Vafilor getting this error when trying to create a Workspace:

image

Vafilor:
> @Vafilor getting this error when trying to create a Workspace:
> 
> image

@rushtehrani Added a migration. Please pull and try again.

rushtehrani:
> @rushtehrani Added a migration. Please pull and try again.

That did it

rushtehrani:
@Vafilor Another issue, I think a merge/rebase has gone wrong:
```
Pod "rbac-test-0" is invalid: spec.containers[0].volumeMounts[1].name: Not found: "sys-dshm"
```

Vafilor:
> @Vafilor Another issue, I think a merge/rebase has gone wrong:
> 
> ```
> Pod "rbac-test-0" is invalid: spec.containers[0].volumeMounts[1].name: Not found: "sys-dshm"
> ```

Taking a look.

rushtehrani:
I think it's the [missing volume](https://github.com/onepanelio/core/pull/352/files#diff-0a9e9900eb966689345307cd4ebb74aeL276-L285)

Vafilor:
> I think it's the [missing volume](https://github.com/onepanelio/core/pull/352/files#diff-0a9e9900eb966689345307cd4ebb74aeL276-L285)

That's most likely it - pushed up a fix.

pkg/workflow_execution.go
```diff
@@ -339,6 +341,31 @@ func (c *Client) CreateWorkflowExecution(namespace string, workflow *WorkflowExe
 		return nil, util.NewUserError(codes.NotFound, "Error with getting workflow template.")
 	}
 
+	if runtimeVars != nil && workflowTemplate.ArgoWorkflowTemplate != nil {
```
rushtehrani, line 344: I don't think this section should be here. It's unique to Workspaces only. Other calls to `CreateWorkflowExecution` may never have `"stateful-set-resource"` or `VirtualService`. Can we inject these into the template before it's passed in here?

pkg/workspace.go
```diff
@@ -98,10 +98,21 @@ func injectWorkspaceSystemParameters(namespace string, workspace *Workspace, wor
...
+	if err != nil {
+		return nil, err
+	}
+
+	runtimeVars, err := workspace.WorkspaceTemplate.RuntimeVars(systemConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = c.CreateWorkflowExecution(namespace, &WorkflowExecution{
```
rushtehrani, line 111: @Vafilor in reference to [this comment](https://github.com/onepanelio/core/pull/352/files#r441838842). Ideally this we want to make the override before they're passed into `CreateWorkflowExecution` here.

rushtehrani:
@Vafilor another issue I ran into is that the nodepool changes don't update after 2 or 3 changes, seems like the watcher stops working after a while.

Vafilor:
> @Vafilor another issue I ran into is that the nodepool changes don't update after 2 or 3 changes, seems like the watcher stops working after a while.

That's weird. I'll take a look.

pkg/workflow_execution.go
Vafilor: @rushtehrani I agree, we need to separate the concerns here.

The code above gets the workflow template and that also gets the k8s argo template. At the moment, we can't modify it beforehand because it gets it in this function.

We could change the function so that we pass in the workflow template with the argo contents pre-loaded.
That way it does less work and is more configurable. 

What do you think?

rushtehrani:
@Vafilor another issue is that now `sys-nodepool` and other runtime params aren't showing in Workspace details.

rushtehrani:
@Vafilor nodepool values not updating when config is updated is still an issue. I'm testing by updating `applicationNodePoolOptions` values in system configmap and I don't see the changes in the UI.

pkg/config.go
```diff
@@ -11,7 +11,39 @@ import (
...
+		})
+
+		return nil
+	}
+
+	return &value
+}
+
+// OnepanelDomain gets the ONEPANEL_DOMAIN value, or nil.
+func (s SystemConfig) OnepanelDomain() *string {
```
rushtehrani, line 34: Let's just call this `Domain`. I think we should also have `FQDN` here too.

Vafilor: @rushtehrani Makes sense, and sure thing. I'm slowly adding them as I encounter them in the code.

pkg/workflow_execution.go
rushtehrani: @Vafilor parameters are showing and configuration reload is working. Is this the only change that remains?

Vafilor: @rushtehrani Yes, I should have it ready by tomorrow morning.

PR #357, opened 2020-06-17
fix: reference to parameters can now be defined as `workspace.parameters`

rushtehrani:
**What this PR does**:
Allows users to reference parameters in Workspace Templates like so:

```yaml
‘{{workspace.parameters.param-name}}’
```

### DIFF
--- a/pkg/workspace_template.go
+++ b/pkg/workspace_template.go
@@ -19,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"net/http"
 	"sigs.k8s.io/yaml"
+	"strings"
 )
 
 func parseWorkspaceSpec(template string) (spec *WorkspaceSpec, err error) {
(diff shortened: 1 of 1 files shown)


pkg/workspace_template.go
```diff
@@ -698,6 +699,9 @@ func (c *Client) generateWorkspaceTemplateWorkflowTemplate(workspaceTemplate *Wo
 		return nil, err
 	}
 
+	workflowTemplateManifest = strings.NewReplacer(
```
Vafilor, line 702: This looks good. 

Any reason for using this instead of `strings.ReplaceAll`?

rushtehrani: > Any reason for using this instead of `strings.ReplaceAll`?

Good question, based on the following references:

https://pkg.go.dev/github.com/metakeule/replacer?tab=doc#pkg-overview
https://medium.com/@vikram.ingawale91/golang-strings-replace-vs-strings-replacer-a7b2d2b71593

It seems that it's more memory efficient and generally faster than the alternatives.

Vafilor: Sounds good to me!

Issue #362, opened 2020-06-18
Add API to return Node pools


rushtehrani:
This can be used in CVAT and other Workspaces. Can we leverage the Workspace template parameters here?

PR #369, opened 2020-06-23
feat: added endpoint to list node pools

Vafilor:
**What this PR does**:

Adds an endpoint to list NodePools.

Fixes onepanelio/core#362


server/workspace_server.go
```diff
@@ -261,3 +261,39 @@ func (s *WorkspaceServer) DeleteWorkspace(ctx context.Context, req *api.DeleteWo
 
 	return &empty.Empty{}, err
 }
+
+// ListNodePools returns a list of the node pools as defined in the system config
+func (s *WorkspaceServer) ListNodePools(ctx context.Context, req *api.ListNodePoolRequest) (*api.ListNodePoolResponse, error) {
+	client := getClient(ctx)
```
rushtehrani, line 267: I think we need an endpoint for config (not namespaced) which will return all system config information that is not a secret. We'll have an equivalent one for namespace configs later too.

Vafilor: @rushtehrani would this just be the default namespace configmap?

rushtehrani: > would this just be the default namespace configmap?

No, this is in `onepanel` namespace.

pkg/config.go
```diff
@@ -33,6 +33,22 @@ func (s SystemConfig) GetValue(name string) *string {
 	return &value
 }
 
+// GetValueOrEmpty returns the value in the underlying map if it exists, otherwise empty string is returned
+func (s SystemConfig) GetValueOrEmpty(name string) string {
+	value, ok := s[name]
```
rushtehrani, line 38: Wouldn't this already return the zero value here if the key doesn't exist?

rushtehrani: Also, should we actually log this? Seems like it'd be unneeded chatter.

Vafilor: @rushtehrani 

1. Yep, yes it would. Will fix.

2. I'm fine with not logging it. Doesn't make sense for this method. What do you think about logging for the `GetValue` method above?

Issue #361, opened 2020-06-18
Workspace names aren't validating correctly


rushtehrani:
For example "600s" was allowed as a name

PR #370, opened 2020-06-23
fix: validate workspace names

Vafilor:
**What this PR does**:

Adds a validation check to workspace names

Fixes onepanelio/core#361

This PR should be tested along with https://github.com/onepanelio/core-ui/pull/133

### DIFF
--- a/pkg/config.go
+++ b/pkg/config.go
@@ -8,7 +8,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"regexp"
 	"sigs.k8s.io/yaml"
 	"strings"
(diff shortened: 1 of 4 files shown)


pkg/config.go
```diff
@@ -88,6 +89,25 @@ func (s SystemConfig) DatabaseDriverName() *string {
 	return s.GetValue("databaseDriverName")
 }
 
+// IsSubdomainValid checks if the subdomain is valid.
+// Nil is returned if it is, otherwise an error
+// message describing the violation is returned.
+func (s SystemConfig) IsSubdomainValid(namespace, subdomain string) error {
```
rushtehrani, line 95: This should be in [pkg/util/validate.go](https://github.com/onepanelio/core/blob/master/pkg/util/validate/validate.go). There is already an unused method in there that may be useful.

PR #374, opened 2020-06-25
feat: setup unit tests with database

Vafilor:
**What this PR does**:

Sets up unit testing with a live database.

Fixes onepanelio/core#

run ```make test-run``` to run the tests.


pkg/workflow_execution.go
```diff
@@ -437,32 +449,29 @@ func (c *Client) insertPreWorkflowExecutionStatistic(namespace, name string, wor
...
-		"created_at":                   createdAt.UTC(),
-		"phase":                        wfv1.NodePending,
-		"parameters":                   string(parametersJSON),
-		"is_archived":                  false,
-	}
-
 	err = sb.Insert("workflow_executions").
-		SetMap(insertMap).
+		SetMap(sq.Eq{
+			"UID":                          UID,
```
rushtehrani, line 454: Lower case `uid` is ok as long as it's consistent across the system. If you're making this uppercase here, then you'd have to it [here](https://github.com/onepanelio/core/pull/374/files#diff-2df54c8de22b7ec30c64350c5cbfd591R536) and everywhere else too.

pkg/workspace_test.go
```diff
@@ -1,30 +1,314 @@
...
+	assert.Equal(t, fm["paused_at"], pq.NullTime{})
+	assert.Equal(t, fm["started_at"], pq.NullTime{})
+
+	terminated := fm["terminated_at"].(time.Time)
+
+	assert.True(t, terminated.After(time.Time{}))
+}
+
+func Test_WorkspaceStatusToFieldMap(t *testing.T) {
+	testWorkspaceStatusToFieldMapLaunching(t)
```
rushtehrani, line 59: Can you elaborate on the prefixes in test function names? I see:

`test`
`Test`
`Test_`
`testClient` (No `_` here as opposed to the public funcs)
`TestClient_`

I am assuming `TestClient_` are tests for methods on instances of `DefaultTestClient()`, and methods prefixed with `testClient` are for the private methods. But then looking at this [line](https://github.com/onepanelio/core/pull/374/files#diff-fcca39e329f7cf9707b4269a7c7e40a7R266), that doesn't seem to be the case? Also should we consistently use `_` for both public and private methods?

Vafilor: General form is `Test[<Struct>]_<Method>` where <struct> is optional. This format comes from Goland - if you type in func Test.... it'll provide auto complete options to this.

`TestClient_` are for methods on `Client` struct, which is a majority of the methods in `pkg`.
So we have `TestClient_<method_name>`. Methods like these are for direct tests of a method, they should completely test a method. 

To test specific cases, you have `testClient` methods. Format is `testClient<MethodName><Condition>`
e.g. ```testClientGetWorkflowTemplateDBEmpty``` tests the `GetWorkflowTemplateDB` method, when the there is no template - empty.

`TestClientGetWorkflowTemplateDB` would call this as a child method. 

```go
TestClientGetWorfklowTemplateDB(...) {
   testClientGetWorkflowTemplateDBEmpty(...)
}
```

In the case where we have two methods where the only name difference is case, like
`GetWorkflowTemplate` vs `getWorkflowTemplate` I've used the `private` word.

```testClientPrivateGetWorkflowTemplate```
vs
```testClientGetWorkflowTemplate``` which is for the exported member.

File: pkg/workspace_template_test.go [hunk 3]
```diff
@@ -52,6 +52,75 @@ routes:
...
+#       image: technosophos/slack-notify
+#       args:
+#       - SLACK_USERNAME=onepanel SLACK_TITLE="Your workspace is ready" SLACK_ICON=https://www.gravatar.com/avatar/5c4478592fe00878f62f0027be59c1bd SLACK_MESSAGE="Your workspace is now running" ./slack-notify
+#       command:
+#       - sh
+#       - -c
+`
 )
 
 func TestParseWorkspaceSpec(t *testing.T) {
```
rushtehrani, line 126: Should this be updated to match the new convention?

Vafilor: Yes, I'll update it.

rushtehrani:
Getting this error when running this branch:

```FATA[0006] Failed to run database go migrations: failed to run Go migration "20200525160514_add_jupyter_workspace_template.go": Workspace template with the name 'JupyterLab' already exists 
Exiting.
```

I have already ran this migration so I'm not sure why it is trying it again.

Vafilor:
> Getting this error when running this branch:
> 
> ```
> Exiting.
> ```
> 
> I have already ran this migration so I'm not sure why it is trying it again.

Are you running it in test mode, or just general build mode?

rushtehrani:
> Are you running it in test mode, or just general build mode?

Running in debug mode - not testing.

rushtehrani:
When a test fails, the test container is not stopped and I get this error when trying to run tests:
```
 $ make test-run
make test-start
docker run --rm --name test-onepanel-postgres -p 5432:5432 -e POSTGRES_USER=admin -e POSTGRES_PASSWORD=tester -e POSTGRES_DB=onepanel -d  postgres:12.3
docker: Error response from daemon: Conflict. The container name "/test-onepanel-postgres" is already in use by container "f30a3914dcdf324348b13ff48bfe15cc41a7fde5c551a95d48c33e5234b2f4ce". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
make[1]: *** [test-start] Error 125
make: *** [test-run] Error 2
```

Issue #331, opened 2020-06-05
Native support for GCS artifact repository


rushtehrani:
Support [GCS object storage](https://github.com/argoproj/argo/blob/master/docs/configure-artifact-repository.md)

Docs and params.yaml comments need to also be updated accordingly.

aleksandrmelnikov:
When testing the in-progress code.
- Need to pull down manifests repo, core, and cli.
- All use the same `feat/core.331-support.gcs.artifact[..]` branch

The CLI repository relies on a library from core.
So, make sure to pull down the core branch first.

Then, in CLI repo, in `go.mod`, add this line
```
module github.com/onepanelio/cli

go 1.13

# This one, replace it with your relevant path.
replace github.com/onepanelio/core v0.11.0 => /home/aleksandr/Code/onepanelio/core
```

aleksandrmelnikov:
### Volume Binding Issues
Ensure your kubectl client and kubectl server versions match: 
`kubectl version --short=true`, if not, change your client to match your server.
Check that the storageClass of the PVC and PV match: 
`kubectl get pv` and `kubectl get pvc -n <namespace>`

Edit the storage-provisioner image from 1.8.1, or similiar, to “latest”
![Pasted image](https://user-images.githubusercontent.com/10407733/87360621-66592e00-c51f-11ea-8440-407f8e948302.png)

Then add the following…
`kubectl edit clusterrole -n kube-system system:persistent-volume-provisioner`
And add new permissions there
```yaml
- apiGroups:
  - ""
  resources:
  - endpoints
  verbs:
  - get
  - list
  - watch
  - create
  - patch
  - update
```

aleksandrmelnikov:
To test inline artifact setting with GCS / S3, this is an example for GCS.
```yaml
entrypoint: main
arguments:
  parameters:
    - name: source
      value: 'https://github.com/onepanelio/tensorflow-examples.git'
    - name: command
      value: python mnist/main.py --epochs=5
volumeClaimTemplates:
  - metadata:
      name: data
    spec:
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 2Gi
  - metadata:
      name: output
    spec:
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 2Gi
templates:
  - name: main
    dag:
      tasks:
        - name: train-model
          template: pytorch
        - name: test
          template: test
          dependencies:
            - train-model
  - name: pytorch
    inputs:
[...]

PR #400, opened 2020-07-10
feat: core.331 support.gcs.artifact.repository

aleksandrmelnikov:
**What this PR does**:
Updated code to support Google cloud storage if available.

Fixes onepanelio/core#331


go.mod
```diff
@@ -30,10 +30,14 @@ require (
...
 	github.com/stretchr/testify v1.4.0
 	github.com/tmc/grpc-websocket-proxy v0.0.0-20200122045848-3419fae592fc
+	golang.org/x/net v0.0.0-20200301022130-244492dfa37a
+	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
+	google.golang.org/api v0.20.0
 	google.golang.org/genproto v0.0.0-20200430143042-b979b6f78d84
 	google.golang.org/grpc v1.28.0
 	google.golang.org/protobuf v1.22.0
 	gopkg.in/yaml.v2 v2.2.8
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
```
rushtehrani, line 40: Why do we need this too? Now we have are using 4 different libraries. Can you use `"sigs.k8s.io/yaml"` instead? I'd like to eventually refactor everything to use that.

pkg/workflow_execution.go
```diff
@@ -889,30 +933,61 @@ func (c *Client) GetWorkflowExecutionMetrics(namespace, uid, podName string) (me
...
+					"UID":       uid,
+					"PodName":   podName,
+					"Error":     err.Error(),
+				}).Error("Metrics do not exist.")
+				return nil, util.NewUserError(codes.NotFound, "Metrics do not exist.")
+			}
+		}
+	case config.ArtifactRepository.GCS != nil:
+		{
+			ctx := context.Background()
```
rushtehrani, line 966: [These lines](https://github.com/onepanelio/core/pull/400/files#diff-2df54c8de22b7ec30c64350c5cbfd591R966-R987) seem to be repeated a bunch of times, can this be put into one method?

go.mod
aleksandrmelnikov: Going to create a separate ticket to investigate this.

- We use yaml.v3 to support reading (and writing) YAML with comments, inline comments, and other metadata.
- We use a pkg/config_types.go in CLI to marshal / unmarshal `params.yaml`
- Same package is used in core to dynamically generate the s3 access and secret keys, for Workflows

I think, ideally, we should use yaml.v3

aleksandrmelnikov: https://github.com/onepanelio/core/issues/409

pkg/workflow_execution.go
aleksandrmelnikov: Refactored

pkg/config_types.go
```diff
@@ -137,15 +138,115 @@ func (s SystemConfig) DatabaseConnection() (driverName, dataSourceName string) {
...
 	Bucket          string
 	Endpoint        string
 	Insecure        bool
 	Region          string
 	AccessKeySecret corev1.SecretKeySelector
 	SecretKeySecret corev1.SecretKeySelector
-	AccessKey       string
-	Secretkey       string
+	AccessKey       string `yaml:"accessKey"`
+	Secretkey       string `yaml:"secretKey"`
```
rushtehrani, line 149: should be `SecretKey`

rushtehrani: Do we need this if we have `ArtifactRepositoryS3Provider`? Can't we just move `AccessKey` and `SecretKey` fields to `ArtifactRepositoryS3Provider` and use that across the board?

pkg/config_types.go
```diff
@@ -162,8 +263,23 @@ func (a *ArtifactRepositoryS3Config) FormatKey(namespace, workflowName, podName
...
+	keyFormat := g.KeyFormat
+
+	keyFormat = strings.Replace(keyFormat, "{{workflow.namespace}}", namespace, -1)
+	keyFormat = strings.Replace(keyFormat, "{{workflow.name}}", workflowName, -1)
+	keyFormat = strings.Replace(keyFormat, "{{pod.name}}", podName, -1)
+
+	return keyFormat
+}
+
 type ArtifactRepositoryConfig struct {
```
rushtehrani, line 280: Can't we just use `ArtifactRepositoryProviderConfig` and make the fields pointers there?

pkg/config_types.go
aleksandrmelnikov: > should be `SecretKey`

Sorry, do you mean the `yaml: "secretKey"`?

rushtehrani: @aleksandrmelnikov the `K` in the actual field name is lower case now, should be uppercase.

aleksandrmelnikov: @rushtehrani I was able to move the attributes from S3Config to S3Provider.

Did you to remove the S3Config struct when you said "use across the board"?

pkg/config_types.go
aleksandrmelnikov: Yes, this makes sense.

pkg/config_types.go
aleksandrmelnikov: Made changes.

Issue #387, opened 2020-07-01
Workspaces - CVAT - Get Workflow Templates via SDK


rushtehrani:
Currently, we're using environment variables, instead we should use the Python SDK's [list_workflow_templates](https://github.com/onepanelio/python-sdk/blob/master/docs/WorkflowTemplateServiceApi.md#list_workflow_templates) API to get a list of templates and only display the ones in CVAT with the label `used-by: cvat`

This will replace the need for `ONEPANEL_MASKRCNN_TEMPLATE_ID` and `ONEPANEL_OD_TEMPLATE_ID` environment variables.

savan77:
@rushtehrani I think we'll have to dynamically generate a form for workflow parameters. Is there an API to get a list of parameters for a given workflow?

rushtehrani:
@savan77 makes sense, there are two parts to this:

1. Get a list of workflow templates to display the available workflows for training. This is the SDK call listed above.
2. Once the user selects one of them, make another SDK call to get the parameters for that  particular WorkflowTemplate. I have to defer to @Vafilor on this one as I don’t recall which API call this is.

savan77:
@rushtehrani To elaborate on the point that we discussed earlier, I was thinking about having a separate menu (just like Tasks, Models, and Analytics) for Onepanel. Maybe we can call it Workflows. There, we can have a `Execute workflow` button which will enlist all available workflows when user clicks on it. Once the user selects a particular workflow, we can get all available parameters for that workflow, assuming we have an API for the same. This is similar to what we have now, but once the workflow has been executed we can show that on that page just like a list of models on the Models page. We can also have a button on that list which links to the executed workflow. 

This will make it easier to keep track of executed workflows and its also generic and not related to model training only. Moreover, this way we will also have a separate component for Onepanel.

aleksandrmelnikov:
Adding the following endpoint:
`/onepanelio/workflow_templates` inside of CVAT.
You will get the following response back:
```json
{
  "page": 1,
  "pages": 1,
  "total_count": 6,
  "count": 6,
  "workflow_templates": [
    {
      "is_archived": null,
      "name": "Test GPU",
      "versions": "11",
      "version": null,
      "created_at": "2020-07-03T01:13:35Z",
      "manifest": null,
      "cron_stats": null,
      "is_latest": null,
      "uid": "test-gpu",
      "modified_at": null,
      "stats": {
        "last_executed": "2020-07-17T20:18:10Z",
        "completed": 3,
[...]

PR #419, opened 2020-07-22
feat: Adding "parameters" to workflow template API response.

aleksandrmelnikov:
**What this PR does**:

Fixes onepanelio/core#387


pkg/workflow_template.go
```diff
@@ -295,6 +295,21 @@ func (c *Client) getWorkflowTemplate(namespace, uid string, version int64) (work
...
+	params, err := workflowTemplate.GetParametersKeyString()
+	if err != nil {
+		return workflowTemplate, err
+	}
+
+	var sParams []Parameter
+	for k, v := range params {
+		p := Parameter{
+			Name:  k,
+			Value: &v,
```
Vafilor, line 307: @aleksandrmelnikov

Vafilor:
@aleksandrmelnikov Let's change the PR name to make it more changelog friendly.

Maybe: feat: Add parameters to workflow template API response 

?

db/go/20200727144157_add_parameters_to_workflow_template_version.go
```diff
@@ -0,0 +1,82 @@
...
+
+	namespaces, err := client.ListOnepanelEnabledNamespaces()
+	if err != nil {
+		return err
+	}
+
+	var pageSize int32
+	var page int32
+	pageSize = 100
+	page = 0
```
Vafilor, line 42: I think you can do 

```go
pageSize := int32(100)
```

Vafilor: @aleksandrmelnikov

db/go/20200727144157_add_parameters_to_workflow_template_version.go
```diff
@@ -0,0 +1,82 @@
...
+	wtsResults := -1
+	for _, namespace := range namespaces {
+		for wtsResults != 0 {
+			wts, err := client.ListWorkflowTemplates(namespace.Name, &paginator)
+			if err != nil {
+				return err
+			}
+			//Exit condition; Check for more results
+			wtsResults = len(wts)
+			if int32(wtsResults) > 0 {
```
Vafilor, line 53: @aleksandrmelnikov Why typecast it to an int32? 

```go
if wtsResults > 0 {
```

Ought to work?

db/go/20200727144157_add_parameters_to_workflow_template_version.go
```diff
@@ -0,0 +1,82 @@
...
+				if err != nil {
+					return err
+				}
+				for _, wtv := range wtvs {
+					params, err := v1.ParseParametersFromManifest([]byte(wtv.Manifest))
+					if err != nil {
+						return err
+					}
+					wtv.Parameters = params
+					err = client.UpdateWorkflowTemplateVersionDB(namespace.Name, wtv)
```
Vafilor, line 69: @aleksandrmelnikov 

Let's add an err check.

```go
err = client.UpdateWorkflowTemplateVersionDB(namespace.Name, wtv)
if err != nil {
   return err
}
```

pkg/workflow_template.go
```diff
@@ -49,6 +55,26 @@ func createWorkflowTemplateVersionDB(runner sq.BaseRunner, workflowTemplateID ui
...
+	if err != nil {
+		return
+	}
+	_, err = sb.Update("workflow_template_versions").
+		SetMap(sq.Eq{
+			"manifest":   wtv.Manifest,
+			"created_at": wtv.CreatedAt,
+			"is_latest":  wtv.IsLatest,
+			"parameters": pj,
+			"version":    wtv.Version,
```
Vafilor, line 70: @aleksandrmelnikov 

I don't think we need to update `created_at`, that shouldn't change.
Same for `version`.

pkg/workflow_template.go
```diff
@@ -498,6 +552,27 @@ func (c *Client) CreateWorkflowTemplateVersion(namespace string, workflowTemplat
...
+	}
+	defer tx.Rollback()
+
+	err = updateWorkflowTemplateVersionDB(tx, wtv)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
Vafilor, line 574: @aleksandrmelnikov 

Looks like namespace is not being used, can we delete it from arguments?

Also, since `updateWorkflowTemplateVersionDB` uses a `sq.BaseRunner` you can pass in `c.DB` as the argument instead of tx.

Finally, this may be more of my preference, but if you name the return as 
```go
(err error)
```

You can omit the err check and have
```go
err = updateWorkflowTemplateVersionDB(c.DB, wtv)

return
```

I find it a little cleaner.

pkg/workflow_template.go
```diff
@@ -575,6 +651,23 @@ func (c *Client) ListWorkflowTemplateVersions(namespace, uid string) (workflowTe
 	return
 }
 
+// ListWorkflowTemplateVersionsDB returns all the WorkflowTemplateVersions from the database.
+// This function is a work-around for ListWorkflowTemplateVersions. Once that function is refactored,
+// this function should no longer be necessary and removed.
+// See: https://github.com/onepanelio/core/issues/436
+func (c *Client) ListWorkflowTemplateVersionsDB(namespace, uid string) (workflowTemplateVersions []*WorkflowTemplateVersion, err error) {
```
Vafilor, line 658: @aleksandrmelnikov 

Let's omit the DB suffix here. Ideally we don't export DB specific methods, so this method might be doing more work than just getting DB contents.

db/go/20200727144157_add_parameters_to_workflow_template_version.go
aleksandrmelnikov:
```
		wtvsResults = len(wtvs)
```
len returns an `int`.
I could do some more typecasting?

db/go/20200727144157_add_parameters_to_workflow_template_version.go
aleksandrmelnikov: Agreed

pkg/workflow_template.go
aleksandrmelnikov: Agreed.

db/go/20200727144157_add_parameters_to_workflow_template_version.go
aleksandrmelnikov: I misunderstood, adding change

pkg/workflow_template.go
aleksandrmelnikov: Agreed

pkg/workflow_template.go
aleksandrmelnikov: Agreed

pkg/workflow_template.go
```diff
@@ -334,6 +388,22 @@ func (c *Client) listWorkflowTemplateVersions(namespace, uid string) (workflowTe
 	return
 }
 
+func (c *Client) listWorkflowTemplateVersionsDBAll(paginator *pagination.PaginationRequest) (workflowTemplateVersions []*WorkflowTemplateVersion, err error) {
+	return c.selectAllWorkflowTemplateVersionsDB(paginator)
+}
```
Vafilor, line 393: @aleksandrmelnikov Do we need this method? Why not just call 
```golang
selectAllWorkflowTemplateVersionsDB
```

pkg/workflow_template.go
```diff
@@ -334,6 +388,22 @@ func (c *Client) listWorkflowTemplateVersions(namespace, uid string) (workflowTe
...
+// See https://github.com/onepanelio/core/issues/436
+func (c *Client) listWorkflowTemplateVersionsDB(namespace, uid string) (workflowTemplateVersions []*WorkflowTemplateVersion, err error) {
+	dbVersions, err := c.selectWorkflowTemplateVersionsDB(namespace, uid)
+	if err != nil {
+		return nil, err
+	}
+
+	return dbVersions, err
+}
+
```
Vafilor, line 406: @aleksandrmelnikov Do we need this method? Why don't we just call
```golang
selectWorkflowTemplateVersionsDB
```

pkg/workflow_template.go
```diff
@@ -575,6 +656,35 @@ func (c *Client) ListWorkflowTemplateVersions(namespace, uid string) (workflowTe
...
+			"Namespace": namespace,
+			"UID":       uid,
+			"Error":     err.Error(),
+		}).Error("Workflow template versions not found.")
+		return nil, util.NewUserError(codes.NotFound, "Workflow template versions not found.")
+	}
+	return
+}
+
+// ListWorkflowTemplateVersionsDBAll returns all WorkflowTemplateVersions with no filtering.
```
Vafilor, line 676: @aleksandrmelnikov typo in the method doc

`ListWorkflowTemplateVersionsDBAll` -> `ListWorkflowTemplateVersionsAll`

pkg/workflow_template.go
aleksandrmelnikov: Agreed

pkg/workflow_template.go
aleksandrmelnikov: Agreed

pkg/workflow_template.go
aleksandrmelnikov: Fixed

aleksandrmelnikov:
Do not merge, found a possible issue with editing workflow templates in the UI.
- Parameters "not null" error.

`pq: null value in column "parameters" violates not-null constraint`

aleksandrmelnikov:
Latest PR is here: https://github.com/onepanelio/core/pull/443
- This had dev merged into it, to get latest changes.

Issue #414, opened 2020-07-15
Can't download output artifacts when the file is large


savan77:
After the workflow has been completed, if I try to download artifacts (i.e logs) of size  ~10MB or higher nothing happens. There is no error or anything else. The file also does not get downloaded.

Here is one for your reference: https://app.cnas-re-uog.onepanel.io/crb/workflows/tf-object-detection-training-44n98.

aleksandrmelnikov:
One issue is too many requests error.
`polyfills-es2015.2987770fde9daa1d8a2e.js:1 GET https://app.cnas-re-uog.onepanel.io/api/apis/v1beta1/crb/workflow_executions/tf-object-detection-training-44n98/artifacts/artifacts%2Fcrb%2Ftf-object-detection-training-44n98%2Ftf-object-detection-training-44n98-984996726%2Fmain.log 429`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

aleksandrmelnikov:
Found this error:
```
{"error":"grpc: received message larger than max (16060619 vs. 4194304)","code":8,"message":"grpc: received message larger than max (16060619 vs. 4194304)"}
```
So it looks like that log file is too large for GRPC to return.

Vafilor:
@aleksandrmelnikov How's this looking? Will it make it into this release?

PR #475, opened 2020-08-04
fix: issue where downloading artifact files from a workflow over 10 MB failed

aleksandrmelnikov:
**What this PR does**:
Increases the server and client message size.
- Both receive and sending.

Fixes onepanelio/core#414

To test, create a workflow that outputs a main.log file of over 10 MB.
Then try downloading it.
You should be able to download the file and not get any console errors in the browser.

### DIFF
--- a/api/api.proto
+++ b/api/api.proto
@@ -19,6 +19,7 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
 	schemes: HTTPS;
 	consumes: "application/json";
 	produces: "application/json";
+	produces: "application/octet-stream";
 	security_definitions: {
 		security: {
 			key: "Bearer";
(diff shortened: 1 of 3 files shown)


api/api.proto
```diff
@@ -19,6 +19,7 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
 	schemes: HTTPS;
 	consumes: "application/json";
 	produces: "application/json";
+	produces: "application/octet-stream";
```
Vafilor, line 22: @aleksandrmelnikov Do we need this?

aleksandrmelnikov: I think yes because when an artifact file is retrieved, core-ui treats it as an "octet-stream".
- And artifact file is returned as a []byte in go.

Adding it here makes notes that our api may return an octet-stream.

Vafilor: @aleksandrmelnikov Alrighty.

Issue #485, opened 2020-08-06
Remove `nvidia-smi` mount


rushtehrani:
We need to validate whether `nvidia-smi` is automatically mounted in Azure and [remove this line](https://github.com/onepanelio/core/pull/393/files#diff-2df54c8de22b7ec30c64350c5cbfd591R265).

Edit: It looks like this has indeed been [fixed](https://github.com/Azure/AKS/issues/1271#issuecomment-662976762) in AKS

PR #486, opened 2020-08-06
chore: Remove nvidia-smi mount

rushtehrani:
**What this PR does**:
Removes the `nvidia-smi` mount.

Fixes onepanelio/core#485

Here's a Workflow that can be run to validate this:
```yaml
arguments:
  parameters:
  - name: command
    value: 'nvidia-smi'
  - name: sys-node-pool
    value: Standard_NC6
entrypoint: main
templates:
- dag:
    tasks:
    - name: train-model
      template: tensorflow
  name: main
- name: tensorflow
  nodeSelector:
    beta.kubernetes.io/instance-type: '{{workflow.parameters.sys-node-pool}}'
  container:
    args:
    - '{{workflow.parameters.command}}'
    command:
    - sh
    - -c
    image: tensorflow/tensorflow:1.13.1-gpu-py3
```

### DIFF
--- a/pkg/workflow_execution.go
+++ b/pkg/workflow_execution.go
@@ -138,32 +138,6 @@ func injectArtifactRepositoryConfig(artifact *wfv1.Artifact, namespaceConfig *Na
 	}
 }
 
-// injectNvidiaGPUFields adds GPU specific fields if there is a GPU request
-func injectNvidiaGPUFields(template *wfv1.Template, systemConfig SystemConfig) {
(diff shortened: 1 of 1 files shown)


Vafilor:
@rushtehrani For this one, if you run that template without the fix - it won't work? It'll only work with the fix?

rushtehrani:
@Vafilor You should get an error without this change now. I believe this is only the case with all new GPU nodes launched.

Issue #365, opened 2020-06-22
Add an `exitHandler` to update Workspace state to failed when an action fails.


rushtehrani:
If a Workflow for creating/resuming/pausing/deleting a Workspace fails, add an exitHandler to update Workspace state to failed and allow users to retry last action that failed.

We also need to look into curl backoff and retry.

PR #569, opened 2020-09-10
feat: added retry last action support for workspaces

Vafilor:
**What this PR does**:

Adds an exit handler so workspaces can better detect failures and adds an endpoint to retry the last action.

Fixes onepanelio/core#365

To test this, I created a new workspace template. Then I manually modified it in k8s to include a `fail` template that is a python script that will fail 50% of the time. With this, you can retry all of the actions.

This template should depend on the first `service` template and the virtual service after should now depend on it too.

```python
- name: fail
  script:
    command:
      - python
      - '-u'
    image: 'python:3.7-alpine'
    name: ''
    resources: {}
    source: |
      import sys

      if random.random() > 0.5:
        sys.exit(-20)
```


api/api.swagger.json
```diff
@@ -3,7 +3,7 @@
   "info": {
     "title": "Onepanel",
     "description": "Onepanel API",
-    "version": "0.12.0-rc.0",
+    "version": "0.12.1",
```
rushtehrani, line 6: I believe this should be 0.13.0

rushtehrani: @Vafilor I think this is the only issue as far as code review.

Issue #581, opened 2020-09-16
Use pod anti-affinity to ensure pods get their own node for resource use


aleksandrmelnikov:
We want to ensure pods get their own nodes.

Pod anti-affinity(1) will utilize the cluster-scheduler to ensure that a node has only one pod, per labelSelector and other logic.

This works for us because we want a workspace to have a node to itself, so it can use all the resources.

Previous work here: https://github.com/onepanelio/core/issues/465

1 - https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

PR #582, opened 2020-09-17
feat: Ensure workspaces get a node to themselves with pod anti-affinity.

aleksandrmelnikov:
**What this PR does**:
When you run core with this functionality, any scheduled workspaces will get a node to themselves.
If you have CVAT on one node, and spin up another CVAT workspace, it'll go to another node.

This works for GPU nodes as well.

Fixes onepanelio/core#581

Scheduling may take a bit longer. Azure was a bit slow for me.


pkg/workspace.go
```diff
@@ -3,15 +3,19 @@ package v1
...
 	"github.com/asaskevich/govalidator"
+	"github.com/ghodss/yaml"
 	"github.com/lib/pq"
 	"github.com/onepanelio/core/pkg/util"
 	"github.com/onepanelio/core/pkg/util/pagination"
 	"github.com/onepanelio/core/pkg/util/ptr"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
+	v1 "k8s.io/api/core/v1"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
rushtehrani, line 18: I'd call this `metav1`

aleksandrmelnikov: Agreed, made change.

Issue #583, opened 2020-09-17
Workflows should use pod anti-affinity


rushtehrani:
Similar to #581, we need to use pod anti-affinity instead of resource requests and limits when a `nodeSelector` is added to a Workflow.

Need to update this [function](https://github.com/onepanelio/core/blob/master/pkg/workflow_execution.go#L143).

PR #585, opened 2020-09-18
feat: Updated Workflows, and Cron created Workflows, to ensure they get their own pod by using Pod AntiAffinity. This replaces resource limits and/or requests.

aleksandrmelnikov:
**What this PR does**:
This PR will ensure that only one workflow will be scheduled on a Node.
This way, that workflow will get the entire Node resources to execute.
- All the templates used by the workflow will get a dedicated Node. So multiple containers will get their own Node.

Fixes onepanelio/core#583

The changes here may not exclude the situation where a Workspace shares a node with a Workflow.
Should I adjust the code to ensure a workspace does not share a node with a workflow?

### DIFF
--- a/pkg/workflow_execution.go
+++ b/pkg/workflow_execution.go
@@ -371,25 +371,18 @@ func ensureWorkflowRunsOnDedicatedNode(wf *wfv1.Workflow, config SystemConfig) (
 			continue
 		}
 
-		var value string
-		for k, v := range template.NodeSelector {
-			if k == *config.NodePoolLabel() {
-				value = v
-				break
-			}
-		}
-		if value == "" {
-			continue
-		}
-		if strings.Contains(value, "{{workflow.") {
-			parts := strings.Split(strings.Replace(value, "}}", "", -1), ".")
-			paramName := parts[len(parts)-1]
-			for _, param := range wf.Spec.Arguments.Parameters {
-				if param.Name == paramName && param.Value != nil {
-					nodeSelectorVal = *param.Value
-					break
+		for _, v := range template.NodeSelector {
+			if strings.Contains(v, "{{workflow.") {
+				parts := strings.Split(strings.Replace(v, "}}", "", -1), ".")
+				paramName := parts[len(parts)-1]
+				for _, param := range wf.Spec.Arguments.Parameters {
+					if param.Name == paramName && param.Value != nil {
+						nodeSelectorVal = *param.Value
+						break
+					}
 				}
 			}
+			break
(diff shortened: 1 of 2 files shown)


rushtehrani:
@aleksandrmelnikov yes, we need to make sure a Workspace doesn't share a node with the Workflow, otherwise there will be an issue with resource allocation.  For example, the Workspace and Workflow might both try to request a GPU.

pkg/workflow_execution.go
```diff
@@ -401,6 +361,49 @@ func (c *Client) createWorkflow(namespace string, workflowTemplateID uint64, wor
...
+			break
+		}
+		template.Metadata.Labels = map[string]string{antiAffinityLabelKey: nodeSelectorVal}
+		addPodAffinity = true
+	}
+	if addPodAffinity {
+		wf.Spec.Affinity = &corev1.Affinity{
+			PodAntiAffinity: &corev1.PodAntiAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
+					{LabelSelector: &metav1.LabelSelector{
```
rushtehrani, line 394: @aleksandrmelnikov can you clean this up and move it to next line?

pkg/workflow_execution.go
```diff
@@ -401,6 +361,49 @@ func (c *Client) createWorkflow(namespace string, workflowTemplateID uint64, wor
...
+						MatchExpressions: []metav1.LabelSelectorRequirement{
+							{Key: antiAffinityLabelKey, Operator: "In", Values: []string{nodeSelectorVal}},
+						},
+					},
+						TopologyKey: "kubernetes.io/hostname"},
```
rushtehrani, line 399: This also seems to be indented incorrectly. I think `go fmt` should've caught this?

pkg/workflow_execution.go
aleksandrmelnikov: Yes, will do

aleksandrmelnikov: Check now?

pkg/workflow_execution.go
aleksandrmelnikov: Check now?
I've had to move it manually on it's own line for go fmt to do something about it.

Vafilor: @rushtehrani Surprisingly it doesn't. I also tried it and get the exact same result.

But if you change some whitespace Like having `key` in  `{Key:...` go on the next line, it formats it differently.

Issue #623, opened 2020-09-30
Persist pip packages, jupyterlab extensions, and vscode extensions between switching and pausing.


aleksandrmelnikov:
Outline
Additionally, we need the following to persist during pause & workspace shutdown.

- pip install
- vscode extensions
Code server supports --extensions-dir
- jupyterlab extensions

We want to put the functionality into the template.
- Less work
- We can give them a best practices, default template
- We don't need to support it in the API core.

One idea was to allocate a specific volume, to store the data.
- Issue is there is a limitation on how many volumes can be run and attached. So we want to avoid contributing to that limit.

Save the packages to a volume.
- Look at attached volumes. Use the biggest one.
- Tie into postStart, preStop
- postStart: Pull the packages from a volume, into the installation directories
- preStop: Push to volume
- Warn user with a comment about running out of space. And a hint in the volume field.
- Documentation

We don't want to use S3/Object Storage because that is more coding effort to support.

Reference:
https://github.com/onepanelio/core/issues/609#issuecomment-699182224

aleksandrmelnikov:
@rushtehrani I don't think we can implement this in the template yaml, unless I am missing something?
We have to use container lifecycle hooks right? So this has to be added in the API Core?

rushtehrani:
@aleksandrmelnikov I might be missing something but our `WorskspaceSpec` uses K8s' container definition: https://github.com/onepanelio/core/blob/8b4a70d95898a89c3e47dfd3cef6006bd68461b0/pkg/workspace_types.go#L59

aleksandrmelnikov:
Ah, okay, I think I jumped a few steps; let me review.

PR #646, opened 2020-10-08
feat: Add lifecycle hooks to VSCode and Jupyterlab Workspaces to provide persistence of conda, pip, and extensions.

aleksandrmelnikov:
**What this PR does**:
Adds a design document of the implementation.
- Particularly, the background and explanation of the solution.

Adds migrations to add lifecycle hooks into the workspace templates.

Fixes onepanelio/core#623

**Checklist**

Please check if applies

- [X] I have added/updated relevant documentation

Required 

- [X] I accept to release these changes under the Apache 2.0 License


db/go/20201008153033_update_jupyter_lab_template.go
```diff
@@ -0,0 +1,170 @@
...
+	defer client.DB.Close()
+
+	migrationsRan, err := getRanSQLMigrations(client)
+	if err != nil {
+		return err
+	}
+
+	if _, ok := migrationsRan[20201008153033]; ok {
+		return nil
+	}
```
Vafilor, line 152: @aleksandrmelnikov This will never be true - there's no need to check for this.

We use it above because of a backwards-compatibility check.
Before we split the migrations into sql and go - they were all sql.
So some sql migrations became go migrations. The check makes sure older systems don't run a go migration that's already been run as a sql migration.

aleksandrmelnikov:
Please don't merge this yet, need to move the docs to another repo.

db/go/20201008153041_update_vscode_template.go
```diff
@@ -0,0 +1,153 @@
...
+	defer client.DB.Close()
+
+	migrationsRan, err := getRanSQLMigrations(client)
+	if err != nil {
+		return err
+	}
+
+	if _, ok := migrationsRan[20201008153041]; ok {
+		return nil
+	}
```
Vafilor, line 98: @aleksandrmelnikov We can remove this check for new go migrations. It's impossible that they were ran as sql migrations.

db/go/20201008153033_update_jupyter_lab_template.go
```diff
@@ -0,0 +1,170 @@
...
+# Docker containers that are part of the Workspace
+containers:
+- name: jupyterlab-tensorflow
+  image: onepanel/jupyterlab:1.0.1
+  command: ["/bin/bash", "-c", "pip install onepanel-sdk && start.sh jupyter lab --LabApp.token='' --LabApp.allow_remote_access=True --LabApp.allow_origin=\"*\" --LabApp.disable_check_xsrf=True --LabApp.trust_xheaders=True --LabApp.base_url=/ --LabApp.tornado_settings='{\"headers\":{\"Content-Security-Policy\":\"frame-ancestors * \'self\'\"}}' --notebook-dir='/data' --allow-root"]
+  env:
+    - name: tornado
+      value: "'{'headers':{'Content-Security-Policy':\"frame-ancestors\ *\ \'self'\"}}'"
+  args:
+  ports:
```
Vafilor, line 19: @aleksandrmelnikov Might be pasting issues, but I think ports needs to be indented below args?

db/go/20201008153033_update_jupyter_lab_template.go
```diff
@@ -0,0 +1,170 @@
...
+    name: jupyterlab
+  - containerPort: 6006
+    name: tensorboard
+  volumeMounts:
+  - name: data
+    mountPath: /data
+  lifecycle:
+    postStart:
+	  exec:
+	    command:
```
rushtehrani, line 30: There is an extra space after `:`

rushtehrani: @aleksandrmelnikov you can run `git diff master` to see these along with some other extra white spaces.

db/go/20201008153033_update_jupyter_lab_template.go
```diff
@@ -0,0 +1,170 @@
...
+import (
+	"database/sql"
+	uid2 "github.com/onepanelio/core/pkg/util/uid"
+	"github.com/pressly/goose"
+)
+
+const jupyterWorkspaceTemplate4 = `
+# Docker containers that are part of the Workspace
+containers:
+- name: jupyterlab-tensorflow
```
rushtehrani, line 12: I believe this should just be called `jupyterlab`

rushtehrani:
It looks like using this approach for restoring pip/conda packages is adding an extra 10 mins delay (with JupyterLab). Need to verify whether this is because conda is taking a long time to restore or if it's somewhere else.

rushtehrani:
It seems like `conda env update` attempts to update all packages, even if they exist.

There is an open issue for this: https://github.com/conda/conda/issues/10218

aleksandrmelnikov:
Good find on the "update all packages".
I'm wondering if instead of restoring the environment, we can add some logic to compare the packages.
- If it exists, skip. If it doesn't install?

rushtehrani:
> I'm wondering if instead of restoring the environment, we can add some logic to compare the packages.

Possibly, or if there is a way to only export user installed packages...

aleksandrmelnikov:
@rushtehrani Note the change in the migration to jupyterlab. This should fix the long start time for it.

Time command to debug postStart hook:
Jupyterlab
```yaml
condayml="/data/.environment.yml";
          jupytertxt="/data/.jupexported.txt";
          if [ -f "$condayml" ]; then { time conda env update -f $condayml;} 2>> /data/time.txt; fi;
          if [ -f "$jupytertxt" ]; then { time cat $jupytertxt | xargs -n 1 jupyter labextension install --no-build;} 2>> /data/juptime.txt; fi;
```

VScode
```yaml
            condayml="/data/.environment.yml";
            vscodetxt="/data/.vscode-extensions.txt";
            if [ -f "$condayml" ]; then { time conda env update -f $condayml;} 2>> /data/time.txt; fi;
            if [ -f "$vscodetxt" ]; then { time cat $vscodetxt | xargs -n 1 code-server --install-extension;} 2>> /data/juptime.txt; fi;
```

db/go/20201008153033_update_jupyter_lab_template.go
aleksandrmelnikov: My original intent was to have args be blank. I realize now it can be removed completely, so I removed it.

db/go/20201008153033_update_jupyter_lab_template.go
aleksandrmelnikov: Useful command. Whitespaces should be gone now.

aleksandrmelnikov:
All issues should be resolved.

rushtehrani:
Getting the message below after resuming a paused JupyterLab workspace. 

I'm not sure if the `--no-build` flag is the way to go if this is going to happen... I did click "rebuild" in this case and it only took a couple of mins to get the it to rebuild and restart. Is the rebuild triggered by UI any different than the command line version?

image

aleksandrmelnikov:
@rushtehrani I was able to recreate this error and I pushed up a fix.
I tested with the following packages.
```
conda install nbresuse 
jupyter labextension install --no-build jupyterlab-topbar-extension jupyterlab-system-monitor
```
The time to install and build seems to be ~5-6 minutes.
```yaml
root@jup2-0:/data# tail jupbuild.txt 
[LabBuildApp] JupyterLab 2.2.5
[LabBuildApp] Building in /opt/conda/share/jupyter/lab
[LabBuildApp] Building jupyterlab assets (build:prod:minimize)

real	2m37.921s
user	3m25.626s
sys	0m48.704s
root@jup2-0:/data# tail time.txt 

Please update conda by running

    $ conda update -n base conda



real	2m8.134s
user	1m21.518s
sys	0m5.535s
root@jup2-0:/data# 
```

db/go/20201008153033_update_jupyter_lab_template.go
```diff
@@ -34,6 +34,7 @@ containers:
           jupytertxt="/data/.jupexported.txt";
           if [ -f "$condayml" ]; then conda env update -f $condayml; fi;
           if [ -f "$jupytertxt" ]; then cat $jupytertxt | xargs -n 1 jupyter labextension install --no-build; fi;
+          if [ -f "$jupytertxt" ]; then jupyter lab build; fi;
```
rushtehrani, line 37: Why not `&&` this to the previous command? Wouldn't this still run even if the previous command wasn't successful?

```
if [ -f "$jupytertxt" ]; then cat $jupytertxt | xargs -n 1 jupyter labextension install --no-build && jupyter lab build; fi;
```

aleksandrmelnikov: Good point, change made.

rushtehrani:
Based on my testing:

- Launch with no builds: 1m 30s
- Launch with latest template (with builds): ~8m
- Launch with `--minimize=False` [flag set](https://discourse.jupyter.org/t/installing-extensions-extremely-slow/2627/6): ~4m

The last option seems like the best option as far as startup times, however since it doesn't minimize the extension files, it could slow down browser load times.

Question: is there a way to save JupyterLab extensions in a different directory that is mounted vs a direct on OS disk?

aleksandrmelnikov:
Let me take a look at changing the jupyter installation directory.

aleksandrmelnikov:
> > Let me take a look at changing the jupyter installation directory.
> 
> Sounds good, if that is not option, I think our best bet is to set the `--minimize` flag.

@rushtehrani Investigated this a bit.
Jupyter lab supports a `--app-dir` flag.

What this does is dictate where jupyterlab installs it's assets and extensions.

`[LabBuildApp] Building in /opt/conda/share/jupyter/lab`
That's our current default.

We can change that to `/data/jupyterlab`, or any other volume.

If we go ahead and save the extensions and jupyter to the mounted disk, we have to make sure the workspace resumes and uses that directory for jupyterlab.
- Either through --app-dir or through an environment variable JUPYTERLAB_DIR.

I'm testing if JUPYTERLAB_DIR will work on start-up.

But is your intent to save the extensions to that volume on shut-down?
And restore from that volume?

rushtehrani:
@aleksandrmelnikov this is only going to be useful if it allows us to either bypass build or reduce the time it takes to build the extensions.

If it does, then putting it in `/data/.jupyterlab` would probably makes the most sense assuming we can still use `/data` as the working directory.

aleksandrmelnikov:
Per discussion here: https://onepanelio.slack.com/archives/GJ88JMG9E/p1603913216187400

We're going to go ahead and use the --minimize=False flag as a trade-off.
Using volumes will require additional steps and logic to ensure it works properly.

PR #691, opened 2020-10-30
feat: Add support for visualization sidecars

aleksandrmelnikov:
- Specifically, if a sidecar is labeled with "sys-tensorboard-like",
code attempts to create services and virtual services (routes).
- Code will create argo templates that are resource type, where the
manifests are the generated resources.
The templates are then added right before the template with the
sidecar that requires these services and routes.

**What this PR does**:

Fixes onepanelio/core#

**Checklist**

Please check if applies

Required 

### DIFF
--- a/pkg/workflow_execution.go
+++ b/pkg/workflow_execution.go
@@ -423,6 +423,18 @@ func (c *Client) createWorkflow(namespace string, workflowTemplateID uint64, wor
 							msg := fmt.Sprintf("sidecar %s must have at least one port.", s.Name)
 							return nil, util.NewUserError(codes.InvalidArgument, msg)
 						}
+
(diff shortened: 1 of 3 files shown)


pkg/workflow_execution.go
```diff
@@ -408,6 +410,140 @@ func (c *Client) createWorkflow(namespace string, workflowTemplateID uint64, wor
 		return nil, err
 	}
 
+	newTemplateOrder := []wfv1.Template{}
+	for tIdx, t := range wf.Spec.Templates {
+		if t.Metadata.Labels != nil {
+			if sidecar, ok := t.Metadata.Labels["sidecar"]; ok {
+				if sidecar == "sys-tensorboard-like" {
```
rushtehrani, line 417: we should probably call this `sys-visualization-tool` or something similar to be more generic.

PR #719, opened 2020-11-11
feat: Terminate Workflow action will still execute exitHandler tasks

rushtehrani:
**What this PR does**:
Using Argo stop command instead of terminate allows the Workflow to still execute exitHandler tasks.

Fixes onepanelio/core#

**Checklist**

Please check if applies

Required 

### DIFF
--- a/pkg/workflow_execution.go
+++ b/pkg/workflow_execution.go
@@ -423,6 +423,18 @@ func (c *Client) createWorkflow(namespace string, workflowTemplateID uint64, wor
 							msg := fmt.Sprintf("sidecar %s must have at least one port.", s.Name)
 							return nil, util.NewUserError(codes.InvalidArgument, msg)
 						}
+
+						serviceNameUid := "s" + uuid.New().String() + "-" + namespace
+						serviceNameUidDNSCompliant, err := uid2.GenerateUID(serviceNameUid, 63)
+						if err != nil {
+							return nil, util.NewUserError(codes.InvalidArgument, err.Error())
+						}
+						serviceName := serviceNameUidDNSCompliant + ".alex001.onepanel.io" //todo grab domain
+
+						serviceTemplateName := "k8s-service-template-" + uuid.New().String()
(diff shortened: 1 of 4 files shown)


rushtehrani:
This also includes changes in https://github.com/onepanelio/core/pull/691, and should be merged after. Also it seems like it is marking the Workflows as failed in the UI instead of terminated.

pkg/workflow_execution.go
```diff
@@ -914,7 +914,7 @@ func (c *Client) FinishWorkflowExecutionStatisticViaExitHandler(namespace, name
 			"finished_at": time.Now().UTC(),
 			"phase":       phase,
 		}).
-		Where(sq.Eq{"name": name}).
+		Where(sq.Eq{"name": name}, sq.NotEq{"phase": "Terminated"}).
```
aleksandrmelnikov, line 917: @rushtehrani This needs to be changed to
```
		Where(sq.Eq{"name": name}).Where(sq.NotEq{"phase": "Terminated"}).
```
The 2nd parameter is ignored, and the "AND" piece is not added.
That's why it's not working.

aleksandrmelnikov (changes requested):
Also, codacy is reporting a possible issue? I'm not seeing what the issue is though.

pkg/workflow_execution.go
rushtehrani: @aleksandrmelnikov good catch regarding `AND`, changing it to:
```go
               Where(sq.And{
			sq.Eq{"name": name},
			sq.NotEq{"phase": "Terminated"},
		}).
```

rushtehrani: @aleksandrmelnikov issue is resolved, you can test with this image: `onepanel/core:4fec204`

aleksandrmelnikov: Great, testing.

aleksandrmelnikov (approved):

PR #771, opened 2020-12-08
feat: Use env var instead of tty for interactive sidecars

Vafilor:
**What this PR does**:

Fixes onepanelio/core#

**Checklist**

Please check if applies

Required 


pkg/workflow_execution.go
```diff
@@ -466,7 +469,16 @@ func (c *Client) injectAccessForSidecars(namespace string, wf *wfv1.Workflow) ([
 		for si, s := range t.Sidecars {
 			//If TTY is true, sidecar needs to be accessible by HTTP
 			//Otherwise, we skip the sidecar
-			if s.TTY != true {
+
+			hasInjectIstio := false
+			for _, e := range s.Env {
+				if e.Name == "inject-istio" && e.Value == "true" {
```
rushtehrani, line 475: Ideally we should have the env var indicate that this is an interactive sidecar... something like `ONEPANEL_INTERACTIVE_SIDECAR=true`

Vafilor: @rushtehrani Good idea - on it.

db/yaml/maskrcnn/20201208155115.yaml
```diff
@@ -0,0 +1,192 @@
...
+      workingDir: /mnt/src
+    nodeSelector:
+      beta.kubernetes.io/instance-type: '{{workflow.parameters.sys-node-pool}}'
+    sidecars:
+      - name: tensorboard
+        image: tensorflow/tensorflow:2.3.0
+        command: [sh, -c]
+        env:
+          - name: ONEPANEL_INTERACTIVE_SIDECAR
+            value: "true"
```
rushtehrani, line 133: can we use single quotes for these?

Vafilor: Sure, just a min.

db/go/20201208155115_replace_tty_with_env_var_for_tfod.go
```diff
@@ -0,0 +1,38 @@
...
+	)
+}
+
+// Down20201208155115 rolls back the environment variable change
+func Down20201208155115(tx *sql.Tx) error {
+	// This code is executed when the migration is rolled back.
+	return updateWorkflowTemplateManifest(
+		filepath.Join("tfod", "20201130130433.yaml"),
+		tensorflowObjectDetectionWorkflowTemplateName,
+		map[string]string{
```
aleksandrmelnikov, line 34: @Vafilor Just to double check, is this Down migration intentional? It seems to add the same map string as the up string?

Vafilor: @aleksandrmelnikov Yes, those are the labels. The code will set the labels, so if we don't provide anything here it should wipe them out.

db/go/20201208155805_replace_tty_with_env_var_for_maskrcnn.go
```diff
@@ -0,0 +1,38 @@
...
+			"used-by": "cvat",
+		},
+	)
+}
+
+// Down20201208155805 rolls back the environment variable change
+func Down20201208155805(tx *sql.Tx) error {
+	// This code is executed when the migration is rolled back.
+	return updateWorkflowTemplateManifest(
+		filepath.Join("maskrcnn", "20201115145814.yaml"),
```
aleksandrmelnikov, line 32: Same as above, Down migration uses the same used-by?

db/go/20201208155115_replace_tty_with_env_var_for_tfod.go
aleksandrmelnikov: Okay, that makes sense. But I see the same key: value values? For up and down?

aleksandrmelnikov (approved):

Issue #828, opened 2021-01-08
Allow variable references in Workflow and Workspace parameters


rushtehrani:
Example:

```yaml
arguments:
  parameters:
  - name: artifacts-location
    value: artifacts/{{workflow.namespace}}/
```

Only [global variables](https://argoproj.github.io/argo/variables/#global) should be supported.

PR #829, opened 2021-01-09
feat: replace parameter namespace values with runtime value

Vafilor:
**What this PR does**:

Fixes onepanelio/core#828

**Checklist**

Please check if applies

Required 

### DIFF
--- a/pkg/workflow_execution.go
+++ b/pkg/workflow_execution.go
@@ -423,7 +423,6 @@ func (c *Client) createWorkflow(namespace string, workflowTemplateID uint64, wor
 	if opts == nil {
 		opts = &WorkflowExecutionOptions{}
 	}
-
 	if opts.Name != "" {
 		wf.ObjectMeta.Name = opts.Name
 	}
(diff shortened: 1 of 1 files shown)


pkg/workflow_execution.go
```diff
@@ -460,6 +459,20 @@ func (c *Client) createWorkflow(namespace string, workflowTemplateID uint64, wor
...
 	}
 
+	newParameters := make([]wfv1.Parameter, 0)
+	for i := range wf.Spec.Arguments.Parameters {
+		param := wf.Spec.Arguments.Parameters[i]
+		if param.Value != nil {
+			// Strip out whitespace from parameter value
+			formattedValue := strings.Replace(*param.Value, " ", "", -1)
+			*param.Value = strings.Replace(formattedValue, "{{workflow.namespace}}", namespace, -1)
+			*param.Value = strings.Replace(*param.Value, "{{workspace.namespace}}", namespace, -1)
```
rushtehrani, line 469: should this use `formattedValue` too?

Vafilor: @rushtehrani Do you mean the 3rd line there?

```
*param.Value = strings.Replace(*param.Value, "{{workspace.namespace}}", namespace, -1)
```

If so, no. It is correct because we replace workflow.namespace above. If we set it to formattedValue we would clear out those changes.

rushtehrani: Ah got it. Minor: might be cleaner to use `ReplaceAll` - it calls `Replace` with `-1` internally.

PR #892, opened 2021-03-05
fix: fixes stability issue with pns executor

lnguyen:
**What this PR does**: 
This fixes stability issue with pns executor when using microk8s which doesn't use docker as executor anymore. It is recommended to use empty dir for argo https://argoproj.github.io/argo-workflows/empty-dir/

Required 

- [X] I accept to release these changes under the Apache 2.0 License

### DIFF
--- a/pkg/workflow_execution.go
+++ b/pkg/workflow_execution.go
@@ -2,11 +2,19 @@ package v1
 
 import (
 	"bufio"
-	"cloud.google.com/go/storage"
 	"database/sql"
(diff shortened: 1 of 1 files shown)


rushtehrani:
Thanks for the PR @lnguyen, this is a highly useful change. We need to test this with a few workflows that [write metrics](https://docs.onepanel.ai/docs/reference/workflows/metrics) and also update the docs accordingly before merging.

PR #904, opened 2021-03-29
feat: Update code to work better with PNS executor

Vafilor:
**What this PR does**:

Updates migrations to better work with PNS executor related changes to volumes.
Also updates workspace migrations to include updated onepanel filesyncer.

The volume path has been changed to `/mnt/tmp/` instead of`/mnt/out/`.

Fixes onepanelio/core#

**Checklist**

Please check if applies

Required 


db/go/20210323175655_update_templates_for_pns.go
```diff
@@ -0,0 +1,80 @@
...
+		initializedMigrations[20210323175655] = true
+	}
+}
+
+// Up20210323175655 update workflows to support new PNS mode
+func Up20210323175655(tx *sql.Tx) error {
+	// This code is executed when the migration is applied.
+	if err := updateWorkflowTemplateManifest(
+		filepath.Join("workflows", "hyperparameter-tuning", "20210323175655.yaml"),
+		hyperparameterTuningTemplateName,
```
rushtehrani, line 21: We actually need to delete this since it's in starter templates now - docs are updated to reflect this change.

Vafilor: On it.

db/yaml/workspaces/jupyterlab/20210323175655.yaml
```diff
@@ -0,0 +1,101 @@
+containers:
+  - name: jupyterlab
+    image: onepanel/dl:0.17.0
```
rushtehrani, line 3: This should be upgraded to onepanel/dl:0.20.0

db/go/20210323175655_update_templates_for_pns.go
Vafilor: Based on discussion with Rush, we are going to do a conditional delete in case someone has actually used this template.

If they have not, it will be deleted.

db/yaml/workspaces/jupyterlab/20210323175655.yaml
Vafilor: Done.

PR #943, opened 2021-08-10
feat: Inference Service API

Vafilor:
**What this PR does**:

Fixes onepanelio/core#

**Checklist**

Please check if applies

Required 


api/proto/inference_service.proto
```diff
@@ -0,0 +1,92 @@
...
+  string minCpu = 5;
+  string minMemory = 6;
+  string maxCpu = 7;
+  string maxMemory = 8;
+}
+
+message CreateInferenceServiceRequest {
+  string namespace = 1;
+  string name = 2;
+  string transformerImage = 3;
```
rushtehrani, line 68: Is there a reason we are exposing only `transformerImage`? Seems a bit inconsistent.

api/proto/inference_service.proto
```diff
@@ -0,0 +1,92 @@
...
+
+  InferenceServicePredictor predictor = 4;
+  Transformer transformer = 5;
```
rushtehrani, line 71: This should be `InferenceServiceTransformer`.

api/proto/inference_service.proto
Vafilor: @rushtehrani That's for convenience. Otherwise you have to specify a transformer, with containers, with envs.

rushtehrani: @Vafilor got it, edited my original comment before seeing your reply.

Vafilor: @rushtehrani Sure, that works.

